PR #21, opened 2015-11-05
Add NetBSD support.

0-wiz-0:
I think this should be converted to autotools and libtool instead (and I'm willing to help with that), but this is makes it at least build on NetBSD.


veorq:
Thanks! an autotool-based build would help indeed :)

PR #25, opened 2015-11-07
Avoid memcmp() for hash verification

paragonie-scott:
https://cryptocoding.net/index.php/Coding_rules#Problem


veorq:
Thanks! that was on my todo list but you were faster :+1:

PR #27, opened 2015-11-07
Paranoid return checks on malloc.

technion:
I appreciate some of these values are quite small and unlikely to fail, however:

* Some of these values are caller defined and could be as large as the caller provides
* Better safe than sorry
* This makes it consistent with other malloc calls


sneves:
I appreciate the thought---this is the right thing to do---but your patch currently leads to a memory leak if the allocation of `out` fails.

technion:
Great catch - also if just one of the earlier ones fail. Have updated.

PR #30, opened 2015-11-08
Document process to identify required encoding lengths.

technion:
The difficulty with magic numbers is that it isn't always obvious where they came from. Throwing 300 at it was assumedly guesswork, which isn't a totally safe approach (even though it turned out to be excess of requirements).


veorq:
That was necessary, thanks!

PR #40, opened 2015-11-28
Dev

flamewow:


flamewow:
I've added some Python implementation, added raw function for hashing password in argon2.c (to call it from Python). It is only first implementation, then i want to add some check to Python code.

veorq:
Thanks! gonna review

sneves:
A few comments:

 - The copyright notice is repeated at the top of `argon2.c`
 - The function `argon2_hash_2py` seems unnecessary. You could achieve the same result with `argon2_hash(t_cost, m_cost, parallelism, pwd, pwdlen, salt, saltlen, hash, hashlen, NULL, 0, type)`.  
 - Similarly, `type` does not need to be changed to `uint32_t`; the underlying type of an enum in C is `int`, so you can use ctypes's `c_int` for it.
 - The error code `ARGON2_ETERNAL_ERROR` is not necessary, as `ARGON2_MEMORY_ALLOCATION_ERROR` already exists for memory allocation failures.

I don't know much about Python, so I can't comment on `argon2.py`. Maybe it would be useful to move the Python stuff to, e.g., `bindings/python`, assuming such language bindings belong to this project. That way there would be a logical way to add bindings to other languages without the `src` directory becoming a mess.

flamewow:
Thanks guys, i'll fix remarks and ping you.

flamewow:
Fixed. 
Also added error codes to python implementation. Check it:
$ make
$ cd bindings/python/
$ python3 -i argon2.py ( or "$ python -i argon2.py" if you are using python2)
>>> hash("password","some_salt")
b'\xa8&x\xc7\xd9\xc4\x1f\xdf[2\xd9hq\xab\xe5\xb4WV\x89\xca\xa4\xee\xb3\x98\xf1I\xd1\xdaf\xe7=\xfaA\x04\xeb\xe1\xfd\x94)\xad\x84\r\x9ed<8xE\xc3\xd3\xfb\x13\xcbN\xcf\\}\xfd-9\x8b\x07@\xd8\x10\x1a\x83\x05\xd5\xfd\xc4m\x9f\xd7\x81\xdcX\x87\xb2\x02\xa9R\xc1\x9d\xaf6\xbb\x8c\xe1vH+\x07\xc7Y\x80\xb3\xb5\xf8\xba\xbd\x87\xd8\xf5\xea\x1a\x04V&\xf7\xde\x9b\x93\x8dbQ\x91e\xf6\xd6\xa2\xd8G8\xe3\x9a\x03\xf3'

veorq:
`bindings/` missing :-)

flamewow:
No, it's here) check it in my fork  (dev branch) https://github.com/flamewow/phc-winner-argon2/tree/dev/bindings

veorq:
Ah sorry was looking at your master

veorq:
Noticed minor issues:

* `argon2.py` hardcodes the relative path `../../libargon2.so`, but it might be better to 1) point to a default place for shared libs (`$LD_LIBRARY_PATH`?), and 2) use the proper extension (`.dylib` on OS X)

* `hash()` overrides Python's built-in `hash()`, I suggest renaming to `argon2()` or `hash_argon2()`

We should at least fix the second.

flamewow:
As soon as we are calling hash function from argon2 module - it's okay.
Calling of that function will be like:
import argon2
print(**argon2.hash**("password", "some salt"))

flamewow:
Added some checks (check it on OS X please), renamed function for case: "from argon2 import hash".
As default now i'm trying to load lib from $LD_LIBRARY_PATH. Also added some testing features, to test   python you may just run argon2.py and argon2_hash("password", "some salt") will be executed.
Awaiting for your tests on OS X. =)

veorq:
Quick test gives:

$ python -i argon2.py
�qĈ�c}�h��ᝆ1�d62;�Lڱ�@��#�`��7Mjz\��ϱX�1_c
e?}=�)�L``֊����z+�c��S-�E

flamewow:
It's okay. For flexibility i'm returning raw data.

flamewow:
Type:
print(base64.b64encode(argon2_hash("password", "some salt")))


to get base64 encoded hash

flamewow:
Seems like it work on windows.

flamewow:
Guys, are we going to merge pull request? So I can create a distribution package and upload on PYPI, if you do not mind.

veorq:
Yeah gonna merge, just wanted to make another review round. Will do today.

veorq:
As discussed on https://github.com/P-H-C/phc-winner-argon2/issues/47, it's better to leave bindings out of the reference code packager. We'll cite bindings in the README though. Is you code on Pypi?

flamewow:
@veorq:
Yeah, here is the link:
https://pypi.python.org/pypi/argon2

veorq:
Thanks, now listed on https://github.com/P-H-C/phc-winner-argon2#bindings

PR #48, opened 2015-12-22
Fix bindings links

ranisalt:
Accidentally the links and link texts were inverted on Markdown :laughing: additionally I linked to profiles on the user handles.


veorq:
Oh, thanks!

PR #50, opened 2015-12-30
Don't dereference a pointer before checking that it's not NULL

jedisct1:
Harmless here, but it doesn't hurt since `fill_segment()` is supposed to be a no-op with a NULL instance.


sneves:
Thanks, good catch.

PR #60, opened 2016-01-08
makefile: preserve external flags

lucab:
This PR alters Makefile to honour existing CFLAGS and LDFLAGS when building libraries and binaries.
Those flags are typically used for system-wide compilation tweaks.

Signed-off-by: Luca Bruno <lucab@debian.org>


veorq:
Good change, thanks!

PR #68, opened 2016-01-12
Fix/workaround some gcc  extra warnings

mbroz:
If you add some more gcc checks, like mentioned below, a lot of warnings appears.

Some are fixed with patches, some like
  ``comparison of unsigned expression < 0 is always false``
  ``comparison is always false due to limited range of data type``
are in range check code and depends on exact #defines...
(Anyway, in the current setting it is dead code because
checks are equal to uin32_t data type on input...)

I would prefer that code should avoid these warning because later
some real problem could appear and it is easy to overlook a new warning.

Tested with Makefile with these added options:
```
CFLAGS += -Wextra \
 -Wsign-compare \
 -Werror-implicit-function-declaration \
 -Wpointer-arith \
 -Wwrite-strings \
 -Wswitch \
 -Wmissing-format-attribute \
 -Wstrict-aliasing=3 \
 -Winit-self \
 -Wunsafe-loop-optimizations \
 -Wdeclaration-after-statement \
 -Wold-style-definition \
 -Wno-missing-field-initializers \
 -Wno-unused-parameter \
 -Wno-attributes \
 -Wno-long-long \
 -Wall -Wuninitialized \
 -Wno-switch \
 -Wdisabled-optimization \
 -Wwrite-strings \
 -Wpointer-arith \
 -Wbad-function-cast \
 -Wmissing-prototypes \
 -Wmissing-declarations \
 -Wstrict-prototypes \
 -Wnested-externs \
 -Wcomment \
 -Winline \
 -Wcast-align \
 -Wcast-qual \
 -Wredundant-decls
```

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -11,11 +11,9 @@
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
 
-#include <stdint.h>
 #include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
-#include <limits.h>
 
 #include "argon2.h"
 #include "encoding.h"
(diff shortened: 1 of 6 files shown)


veorq:
Thanks!

PR #69, opened 2016-01-12
Introduces an error state for threading failures. Fixes #64.

### LINKED ISSUE #64: exit(-1) on thread creation/join failures
Before anything - thanks for publishing this!

In reading through argon2's source, I noticed three occurences of `exit(-1`) in `core.c`:

https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.c#L289
https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.c#L308
https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.c#L322

...which means programs that link against argon2 might terminate unexpectedly.
Consider, for example, the following Python program that uses [argon2_cffi](https://github.com/hynek/argon2_cffi):

````python
import resource
import sys

from argon2 import PasswordHasher

if sys.argv[1:]:
    resource.setrlimit(resource.RLIMIT_NPROC, (1, 1))


ph = PasswordHasher()
print ph.hash("s3kr3tp4ssw0rd")
print "More code runs."
````

When run on a Linux system with no arguments it runs in its entirety.  However, with additional arguments, the program reduces the allowed number of processes/threads below what argon2 requires, and so argon2 causes an unexpected termination:

````
$ python example.py
$argon2i$m=512,t=2,p=2$DOoJURupbQWaLfR2ye12tQ$kiBvcn+l+6RBRTaUVB7UiQ
More code runs.
$ python example.py 1
ERROR; return code from argon2_thread_create() is 11
````

An API that reserves the right to terminate your program seems troubling, especially as it might enable an attack.  A web server under resource pressure, for example, could be pushed into a cascading failure.

At the very least this behavior should be documented.  I'll be happy to submit a PR either for an API change or for a documentation change.

Thanks again for argon2!

technion:
fill_memory_blocks() configured to return a status, this introduces checks
for a number of a different error conditions, inlcuding calloc() and NULL
pointer failures. These were previously ignored errors.
Thread failures no longer call printf/exit, ensuring better support for
higher level bindings. fill_memory_blocks() is only called in one place,
it's return value is now checked.

Implications for bindings: new error code may need dedicated handling.

This patch can be tested by setting ulimit -u and running ./argon2 with a
relevant thread count.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -102,7 +102,9 @@ static const char *Argon2_ErrorMessage[] = {
     /*},
 {ARGON2_ENCODING_FAIL, */ "Encoding failed",
     /*},
-{ARGON2_DECODING_FAIL, */ "Decoding failed", /*},*/
+{ARGON2_DECODING_FAIL, */ "Decoding failed",
+    /*},
+{ARGON2_THREAD_FAIL */ "Threading failure", /*},*/
 };
 
 
@@ -151,8 +153,11 @@ int argon2_core(argon2_context *context, argon2_type type) {
     }
 
     /* 4. Filling memory */
-    fill_memory_blocks(&instance);
+    result = fill_memory_blocks(&instance);
 
+    if (ARGON2_OK != result) {
+        return result;
+    }
     /* 5. Finalization */
     finalize(context, &instance);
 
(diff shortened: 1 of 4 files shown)


veorq:
Thanks!

PR #71, opened 2016-01-13
docs(readme) add Lua binding

thibaultCha:
Wrote some Lua binding.

Also proposed to add Argon2 to Homebrew: Homebrew/homebrew#48019, we'll see how it goes!


veorq:
Thanks!

PR #79, opened 2016-01-21
Implement verify fix from libsodium

### LINKED ISSUE #77: hash_encoded produces encodings that can't be verified
`hash_encoded` is validating input against `#DEFINE`s in `argon2.h`. However, `decode_string` is using its own validation which is different from these defines. For example, I can generate this hash: `$argon2i$m=20000,t=1,p=475$QW4gKOKIniwx...` - but I can't decode it because `p=475`, but `decode_string` checks:

```c
    /*
     * The parallelism p must be between 1 and 255. The memory cost
     * parameter, expressed in kilobytes, must be at least 8 times
     * the value of p.
     */
    if (ctx->lanes < 1 || ctx->lanes > 255) {
        return 0;
    }
```

I imagine there are more problems in `decode_string` that just the `p` parameter.

technion:
Fixes #77 , and potentially several related issues stemming from inconsistent verifiers. Original patch from Libsodium.

The original patch would cause verifies, and called in run.c, to fail with ARGON2_PWD_PTR_MISMATCH, and then ARGON2_FREE_MEMORY_CBK_NULL - have introduced extra zeroing to correct this.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -297,6 +297,8 @@ int argon2_verify(const char *encoded, const void *pwd, const size_t pwdlen,
     ctx.adlen = 512;
     ctx.saltlen = 512;
     ctx.outlen = 512;
+    ctx.allocate_cbk = NULL;
+    ctx.free_cbk = NULL;
 
     ctx.ad = malloc(ctx.adlen);
     ctx.salt = malloc(ctx.saltlen);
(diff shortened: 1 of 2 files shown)


technion:
This largely conflicts with 56c177a00 - I'm happy to close it, or to update it, can someone let me know if this is desired?

veorq:
I believe it would improve the code, so would be great if you could update :-)

PR #80, opened 2016-01-21
Utilise existing constants more consistently.

technion:
No functional changes - but better use of existing constants reduces risk of future bugs and reads better.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -175,15 +175,15 @@ int argon2_hash(const uint32_t t_cost, const uint32_t m_cost,
 
     /* Detect and reject overflowing sizes */
     /* TODO: This should probably be fixed in the function signature */
-    if (pwdlen > UINT32_MAX) {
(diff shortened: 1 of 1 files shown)


src/argon2.c
```diff
@@ -175,15 +175,15 @@ int argon2_hash(const uint32_t t_cost, const uint32_t m_cost,
 
     /* Detect and reject overflowing sizes */
     /* TODO: This should probably be fixed in the function signature */
```
technion: I left this TODO because it was pre-existing, but I'm not sure there's a valid TODO remaining.

lucab: I guess this refers to the fact that function signature takes `size_t` (whose size is guaranteed >=16bits, but arch specific), but argon2_context and the rest of the code operates on `uint32_t`. There are casts sprinkled around in several places for this.
I think the TODO remark is still valid.

veorq:
Ok to merge this one, thanks!

Issue #55, opened 2016-01-08
Digest sizes < 12 fail to verify


hynek:
Hi,

a user of argon2_cffi found an edge case where argon2 will create a hash but verification will fail with a decoding error:  hynek/argon2_cffi#4

I could narrow it down to digest lengths that are less than 12 bytes (I used your run.c for that, not my bindings).

veorq:
Thanks, we'll investigate

hynek:
Pleas note that he found more such cases.  Feel free to ask questions on the linked ticket. :)

If you'd be interested we could build a test harness in Python for Argon2 or something.

DRMacIver:
I can answer questions here too. :-)

The two other cases that fail in the same way - that is, a hash is created successfully but the resulting hash cannot validate the original password, are when the hash length or salt length are > 512. 

These three cases are the only failures I'm aware of, though it's possible that they're hiding other issues that the test in question may uncover if these are fixed.

technion:
Is this caused by #77  ?

technion:
In terms of progressing this.. does Argon2 actually make sense to support the salt lengths that it does? I'm having visions of someone writing a frontend with salts that are 0xFFFFFFFF in length, and calling it "maximum possible Argon2 strength".

Would setting ARGON2_MAX_SALT_LENGTH to 512 be appropriate?

I appreciate this thread appears to suggest there's a bigger issue behind it - but it would resolve part of this issue.

khovratovich:
Found the bug, working on the patch

On Thu, Jan 21, 2016 at 4:53 AM, technion <notifications@github.com> wrote:

> In terms of progressing this.. does Argon2 actually make sense to support
> the salt lengths that it does? I'm having visions of someone writing a
> frontend with salts that are 0xFFFFFFFF in length, and calling it "maximum
> possible Argon2 strength".
>
> Would setting ARGON2_MAX_SALT_LENGTH to 512 be appropriate?
>
> I appreciate this thread appears to suggest there's a bigger issue behind
> it - but it would resolve part of this issue.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/P-H-C/phc-winner-argon2/issues/55#issuecomment-173443814>
> .
>



-- 
Best regards,
Dmitry Khovratovich

khovratovich:
[...]

PR #81, opened 2016-01-21
Safer handling of initial string sizes.

technion:
This removes the ability to create some hashes that cannot be verified correctly. Addresses many of the issues raised in #55. Note in normal cases, this will reduce allocated memory.


technion:
Closing as it was rendered redundant by 56c177a00.

PR #82, opened 2016-01-22
Remove // style comments.

technion:
C89 flags in the existing Makefile mean, this library does not built with these comments in place.


veorq:
Thanks!

PR #86, opened 2016-01-23
Fix bench and genkat after low-level API renaming

lucab:
This fixes build errors in `make bench` and `make test` after low-level API refactoring in d9b0bafe37dbb94f22748b7bc73268b073e4934c.


daniel-dinu:
I forgot about these. Thanks!

Issue #85, opened 2016-01-23
Continuous Integration


daniel-dinu:
Given the number of supported platforms and that it is hard to test each code change on all supported platforms, I think it might be useful to have a CI system to automatically detect and eventually avoid the compilation errors.

It would be nice to build at least on the following systems:
- Debian/Ubuntu
- Mac OS X
- Windows

Any volunteer?

lucab:
Other github projects I touched use travis-ci (Mac + Linux) and appveyor (Windows) for CI.
I guess another good step is to start introducing a testsuite, to catch regressions earlier.

technion:
Been raised before in #28.

hynek:
+1 for also using AppVeyor because making sure that code compiles on all VisualStudios is rather finicky (cf. #91).  You’ll have to use msinttypes though.

ranisalt:
Now that #89 is merged it'd be nice to CI this repository, and it would also be nice to check code coverage for tests too. I use lcov to check coverage on [node-argon2](https://github.com/ranisalt/node-argon2) ([check here](https://github.com/ranisalt/node-argon2/blob/master/.travis.yml#L20)) but test compilation requires [extra flags to be set](https://github.com/ranisalt/node-argon2/blob/master/binding.gyp#L37-L38).

technion:
Regrettably I'm not familiar with Appveyor - but now that #89 is in I'll be creating a Travis config to cover the Ubuntu/Mac side.

hynek:
Would it be possible to sneak in extra compiler flags?  There are a few that complain about C99isms but might be too modern to be part of the regular build chain.

technion:
@hynek  I think this is the perfect place for such a thing - if you can advise what they are I will add.

hynek:
`-Werror=declaration-after-statement` would be one important one that would have prevented two recent regressions.  I’m sure there are more, but I’m a bit out of the C game (and not regretting it really :))

technion:
@hynek pushed.

hynek:
maybe you could try `-pedantic-errors` too?  not sure what it exactly does tho.  as said, I’m a bit rusty. :)

ranisalt:
Why not go straight to `-Wall -Wextra -Wpedantic` in a test? And people please `make test` before pushing and it'll be fine :)

technion:
Note it's already running -Wall -Wextra -Werror.
[...]

PR #89, opened 2016-01-25
Introduces an automated test suite.

technion:
With respect to working on #85, I'd like to suggest introducing a test suite.

The existing kats/test.sh is great at testing block internals during a hash, but leaves large parts of functionality open. I also note that the majority of recent commits and bugfixes have related to the encoding, decoding, and error management.

This test suite covers all those elements, and should integrate well with a chosen CI system. If this PR is accepted I would be happy to assist further in resolving #85 by introducing a Travis configuration.

This PR also sets CFLAGS on test compilations to utilise additional compiler warnings, and ASAN, which provides further verification against several classes of bugs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,6 @@ libargon2.dylib
 src/*.o
 src/blake2/*.o
 genkat
-test.c
 .idea
 *.pyc
+testcase
--- a/Makefile
+++ b/Makefile
@@ -90,8 +90,11 @@ dist:
 		cd ..; \
 		tar -c --exclude='.??*' -z -f $(DIST)-`date "+%Y%m%d"`.tgz $(DIST)/*
 
-test:
+test:   $(SRC) src/test.c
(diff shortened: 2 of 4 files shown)


lucab:
What about having a `tests` directory, introducin a common test runner and then splitting tests by topic? I found such setups to be easier to navigate and be augmented in the long term.

technion:
I have no problem with this - except to say this library basically has two functions that need testing. It's not really in the same position as, for example, my Ruby bindings, which have a much more in-depth test suite.

khovratovich:
Could you update README accordingly?

technion:
I've updated the README. Let me know if you have any concerns with the content.

I've deliberately left actually setting up CI for a separate PR, to avoid the introduction of a test suite becoming a bikeshed over CI solutions.

khovratovich:
Thanks!

PR #95, opened 2016-01-26
More C99

hynek:
Also marks test as .PHONY otherwise you'll get:

```
make: `test' is up to date.
```

when running "make test".

***

I wish I could give out them all at once to you but currently I’m bound by running it against my CI: 

- https://ci.appveyor.com/project/hynek/argon2-cffi/build/job/svg9lob397q7v4mu
- https://ci.appveyor.com/project/hynek/argon2-cffi/build/job/akhhjb890qf81u2t

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,5 +93,7 @@ dist:
 test:
 		@sh kats/test.sh
 
+.PHONY: test
+
 format:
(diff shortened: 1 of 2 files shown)


hynek:
BTW, these errors could be avoided by setting `-Werror=declaration-after-statement`.  I’m not sure what GCC added it though so it could add other problems…

PR #99, opened 2016-02-02
Travis

technion:
This introduces a configuration for Travis CI, which covers Ubuntu and, for most practical terms, OSX, in relation to #85.

You can see the result of this implementation here:
https://travis-ci.org/technion/phc-winner-argon2

Travis CI will also link the relevant build on every PR, allowing assurance that such pulls don't break the build. Note, a repository owner will still have to visit travis-ci.org and enable the service for this to take effect.

This test runs with ASAN and FORTIFY_SOURCE. This does not reflect (all) deployment configuration, but provides an increased assurance in the safety of the result.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: c
+compiler:
+ - clang
+script: make && make test
+
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ $(LIB_ST): 	$(OBJ)
 clean:
 		rm -f $(RUN) $(BENCH) $(GENKAT)
 		rm -f $(LIB_SH) $(LIB_ST) kat-argon2* 
+		rm -f testcase
 		rm -rf *.dSYM
 		cd src/ && rm -f *.o
 		cd src/blake2/ && rm -f *.o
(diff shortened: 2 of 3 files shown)


thibaultCha:
Very nice :+1:

- Owners could also enable the continuous integration to verify PRs
- Considered that little fancy badge in the README or is it not the cusotm?

technion:
@thibaultCha I would be happy to implement the badge in the README, but if I try to do that now, you will get a broken badge. I would encourage repo owners to accept this PR, then enable Travis, before that discussion.

thibaultCha:
Yes. Tho that could be done before merging this.

What about coverage? I believe gcov + coveralls could be a nice setup even tho there might be better alternatives I a not aware of. 

Also does the clang formatter implement a check, so that another check is done on PRs, maybe by Travis too (even if I think other services specialize in style checking) for the validity of the code style?

technion:
clang-formatter does not produce a "pass" or "fail", but automatically reformats the test in the given configuration. This isn't appropriate for a CI.

This did however remind me to run the formatter over the test suite.

ranisalt:
It would be nice to have a build matrix with both GCC and Clang. Check how it's [done here](https://github.com/ranisalt/forgottenserver/blob/master/.travis.yml#L1-L6).

I'd be happy to add coverage support if you want to.

Issue #74, opened 2016-01-15
enum argon2_error_codes should not be part of the public interface


lucab:
`argon2.h` currently exposes a `enum Argon2_ErrorCodes` and the related `argon2_error_codes` typedef. However, this enum is not used by any public library method, as errors are signaled by returning non-zero integers.

In order to improve privacy (stop apps abusing internal error codes), compatibility (guard code changing across versions) and simplifying versioning, I would suggest to move this enum declaration in a private place and to update the doc not to mention `ARGON2_OK`.

lucab:
Additionally, currently all error codes are positive integers while all methods deal with *signed* integers. This could led to confusion and bugs, as seen on #66. Should all methods be more specific and operate on *unsigned int* instead?

khovratovich:
I am usually in favor of unsigned int wherever it is meant, though there is a point of view that uints lead to bugs when comparing to zero in loops/whiles.

lucab:
I'd limit this report to error codes only (the `int` currently returned by public method and the fed into `error_message()`). I can't imagine a scenario where somebody may want to use them as loop conditions. Other int/uint mismatches exist in the library, but I would address them separately (I saw a TODO in place yesterday). 

Instead, after some further research I saw that some bindings (like the ruby one) already re-export them while most of the other bindings don't. So the options on the table are:
 * leave the enum in place and return an enum instead of an int. Users will need to update/re-build every time the enum change. This make dynamic lib quite uncomfortable, but allows bindings to hardcode a list of errors (which needs to be separately maintained/updated). As a packager I personally dislike this, but developers may think otherwise.
 * make the enum private and move to unsigned int. It makes the library more opaque (no way to constant-match error codes in apps), but allow for better forward compatibility. Additionally you can either return `unsigned int` (arch-specific) or uint16_t (fixed width). I'm a supporter of using fixed width types, but I may understand not everybody love them.

khovratovich:
I support the latter option; feel free to propose a patch. @veorq @sneves counteropinions?

daniel-dinu:
I argue for the first option: 

As a developer who uses the library, I would like to have access to the error codes. I want to know why a function call returned an error and to be able to take some actions for some error codes. 

Once the library is stable, the chances of future changes in the enum are very low.

The other int/uint mismatches you identified are linked to the use of the error codes?

lucab:
> The other int/uint mismatches you identified are linked to the use of the error codes?

[...]

PR #108, opened 2016-02-20
Address issue #74

sneves:
This is my proposed resolution to #74. I'm making this a pull request so it gets feedback.

 - Error codes which do not represent success are now negative.
 - There is no longer an `ARGON2_ERROR_CODES_LENGTH` member that can silently change between API versions.

The only question that remains, in my view, is whether to remove the error codes from the public `argon2.h` header and make them internal (or maybe in an independent public header, e.g., `argon2-error.h`). I lean against it, since it might be useful for a user to be able to discriminate between error types, but either way is fine for me.


khovratovich:
I like the idea, and I am also in favor of leaving the error codes in `argon2.h`

ranisalt:
> Error codes which do not represent success are now negative.

Can you explain why? Please mind that it's an enum and users should never directly compare the return code to a number (e.g. `result < 0`)  but rather compare to a code (e.g. `result != ARGON2_OK`).

SparkDustJoe:
In most other schemes I've used, there is usually a return value
representative of the number of bytes processed or zero for fail... OR...
Zero for success and some 0x80000000 number representative of a specific
error.  Never see mixed positive and negative numbers and zero equals
success (I'm not counting string comparisons).

ranisalt:
Please, it's an enum, although in low level `ARGON2_OK` equals 0 it is NOT semantically the same. It's not like "the function returns a positive number" but rather it returns a symbol, represented by a number with no direct meaning.  I don't even understand why there are values since enum values should be assigned automatically.

sneves:
It is a common idiom in C that functions return `0` for success and `< 0` for errors. This is not the case everywhere; on Windows, it is common that nonzero represents success, and zero is an error. But on Unix this seems to be the norm. `0` is also the default C success value, as shown by the definition of the `main` function, even though `EXIT_SUCCESS` also exists and is not required to equal `0`.

Enums in C are essentially just integers. Unlike in C++, where you can have strong type-safe enums, here they're essentially no better than a long list of `#define`s. Also, note that the return values of all functions are `int`, not `argon2_error_code`. 

Anyway, the point is that you can continue to use enums just the way you did before; nothing's changed about that. The only change is that a common C idiom is supported without breaking things.

ranisalt:
Oh, fine, I do not have the background of a C programmer.

khovratovich:
Please feel free to resolve the conflicts and merge.

PR #109, opened 2016-02-20
Add options for hash length and bare output to exe

alipha:
I saw on https://github.com/khovratovich/Argon2/issues/21 that it was requested to be able to have the argon2 utility be able to output only the hexadecimal of the hash, and I thought that was useful, so I implemented that (-r) along with outputting only your $argon2i$... encoded string (-e). I also thought it might be nice to generate different output hash lengths, so I added that too (-h length). A max length of 1024 seems a little large, but having a max at all seems a bit arbitrary.

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -27,7 +27,8 @@
 #define LOG_M_COST_DEF 12 /* 2^12 = 4 MiB */
 #define LANES_DEF 1
 #define THREADS_DEF 1
-#define OUT_LEN 32
+#define OUT_LEN_DEF 32
+#define MAX_OUT_LEN 1024
 #define SALT_LEN 16
 /* Sample encode:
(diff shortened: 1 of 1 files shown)


ranisalt:
While you're at it, mind that you need not (and IMHO should not) to hardcode an ENCODED_LEN, as it can be dynamically calculated, [check this function](https://github.com/ranisalt/node-argon2/blob/e804fed9483173d0a2d04c88430d81f7bc9e54ac/src/argon2_node.cpp#L20-L41).

As of today, the utility only permits salts up to 16 bytes and your solution keeps an upper limit, although increasing it.

khovratovich:
Could one of you bring the encoded calculation function to the code? Then we can approve this PR.

ranisalt:
Gonna do it as soon as I get home.

khovratovich:
Please resolve the conflicts.

daniel-dinu:
@alipha please resolve the conflicts, else your pull request will be closed

alipha:
My apologies for the lack of attention of this pull request.

PR #111, opened 2016-02-22
Issue different error codes for verify failure and mismatch

ranisalt:
Currently the `argon2_verify` function returns `ARGON2_DECODING_FAIL` for both if there was any error during the hashing (even if it was not a decoding failure) or if the password does not match with the supplied hash.

This commit makes a mismatching hash return `ARGON2_VERIFY_MISMATCH` instead, making it possible to detect if there was a failure or if the password just does not match. Additionally, it forwards the verification hashing error for better diagnostic.

There's also a test for mismatching password verification.

### DIFF
--- a/include/argon2.h
+++ b/include/argon2.h
@@ -140,6 +140,8 @@ typedef enum Argon2_ErrorCodes {
 
     ARGON2_DECODING_LENGTH_FAIL = 34,
 
+    ARGON2_VERIFY_MISMATCH = 35,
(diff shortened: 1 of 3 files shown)


ranisalt:
The appveyor integration will never work because there's no appveyor file, could it be disabled meanwhile?

veorq:
Good addition, thanks. We'll fix the Appveyor issue

Issue #112, opened 2016-02-28
Optimisations fail to build on powerpc64le


daxtens:
Hi,

I've tried to build argon2 with optimisations on powerpc64le, and I am hitting a compile failure:
```
dja@dja-builder ~/f/phc-winner-argon2> uname -a
Linux dja-builder 4.2.0-30-generic #35-Ubuntu SMP Fri Feb 19 13:50:54 UTC 2016 ppc64le ppc64le ppc64le GNU/Linux
dja@dja-builder ~/f/phc-winner-argon2> env OPT=TRUE make
rm -f argon2 bench genkat
rm -f libargon2.so libargon2.a kat-argon2* 
rm -f testcase
rm -rf *.dSYM
cd src/ && rm -f *.o
cd src/blake2/ && rm -f *.o
cd kats/ &&  rm -f kat-* diff* run_* make_*
cc -std=c89 -pthread -O3 -Wall -g -Iinclude -Isrc -march=native  src/argon2.c src/core.c src/blake2/blake2b.c src/thread.c src/encoding.c src/opt.c src/run.c -o argon2
cc: error: unrecognized command line option ‘-march=native’
Makefile:75: recipe for target 'argon2' failed
make: *** [argon2] Error 1
```

This is due to powerpc gcc not supporting `-march`. We do have `-mcpu` and `-mtune`, which could be used instead. However, if I change `-march` to `-mcpu`, I get other problems:

```
In file included from src/opt.c:19:0:
src/opt.h:18:23: fatal error: emmintrin.h: No such file or directory
compilation terminated.
```

It looks like you're using Intel intrinsics, meaning that your optimised code would only ever compile on Intel. Is that right?

If so, would it be possible to make `OPT=TRUE` behave differently on platforms other than x86? For example:
  * x86: `-march` and intrinsics
  * other platforms supporting `-march`: `-march` and reference code
  * other platforms supporting `-mcpu`: `-mcpu` and reference code
  * other platforms supporting neither: reference code only

I'm happy to help with compiler flag detection macros or anything else if that would help.

sneves:
You are correct, the optimized version is x86-specific.

ranisalt:
It would be a piece of cake using a different build system as they often have tools to check for compiler options (e.g. CMake's [check_c_compiler_flag](https://cmake.org/cmake/help/v3.0/module/CheckCCompilerFlag.html)) but not so using plain Makefiles.

technion:
[...]

PR #113, opened 2016-02-29
Utilise OPT on supported platforms.

technion:
There are currently a number of issues with the OPT parameter.

- As far as I can see, it's undocumented. I only know about it because I did some work in the Makefile.
- The fact it just refers to "optimisations" and not "x86 specific optimisations" can be confusing
- Releases and distributions are forced to either utilise the flag, and write apps that are x86 specific, or not use the flag, and release suboptimal code.
- This led to #112.
- The majority of bindings have utilised the path of least resistance and not used this flag at all
- This means the majority of deployments are not using the optimised path
- Slower code may lead to reduced work strengths than servers are suited for

This change aims to work around it. The logic looks backwards - but I've tested on EC2 (x86_64) and Raspberry Pi (armv71) and it runs correctly. Maybe a few +1's on different platforms would go down well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,12 +23,13 @@ CFLAGS += -std=c89 -pthread -O3 -Wall -g -Iinclude -Isrc
 CI_CFLAGS := $(CFLAGS) -Werror=declaration-after-statement -D_FORTIFY_SOURCE=2 \
 				-Wextra -Wno-type-limits -Werror -coverage
 
-#OPT=TRUE
-ifeq ($(OPT), TRUE)
+OPTTEST := $(shell $(CC) -Iinclude -Isrc -march=native src/opt.c -c 2>/dev/null; echo $$?)
+# Detect compatible platform
+ifneq ($(OPTTEST), 0)
+	SRC += src/ref.c
+else
 	CFLAGS += -march=native
 	SRC += src/opt.c
-else
-	SRC += src/ref.c
 endif
 
 BUILD_PATH := $(shell pwd)


daxtens:
Confirming that `make; make test` succeeds on powerpc64le. +1 and many thanks from me.

veorq:
```
+#Will match i386 and x86_64
 +ifeq (,$(findstring 86, $(ARCH)))
```
A bit worried that this might match other architectures and have unwanted effects, though can't think of other archs at the moment.

technion:
As far as I can determine, you can't trivially write an OR condition to match exactly either complete string, which complicated the solution. I've often referred to this curl file as listing every architecture curl supports (which is everything that makes sense to support) and can't see a clash:

https://github.com/curl/curl/blob/093f9cd38de68f1a327ac8352cfb1168c1140765/include/curl/curlbuild.h.dist

ranisalt:
> +#Will match i386 and x86_64

It could be reworded, I've seen some places where `uname -m` outputs `i586` or `i686`.

sneves:
This check misses at least `amd64`, which is the name used used in the BSDs. There's a big list of possible machine names at [Wikipedia](https://en.wikipedia.org/wiki/Uname). Notice that `uname -m` is supported everywhere, but `uname -p` is not.

ranisalt:
> This check misses at least amd64

In addition, it's not possible to check for "86 or 64" since it would match sparc64

technion:
@sneves Good point - will test out alternatives such as @ranisalt's suggestion.

I do believe it's important to note that a false optimal compile can be breaking, a false suboptimal isn't.

technion:
@ranisalt  I don't know if I'm using it wrong but

    ifeq (,$(findstring (?:amd|x86_)64|[ix]\d?86(?:pc|-AT386)?, $(ARCH)))

Never matches on my x86_x64.

technion:
After reviewing the list supplied by @sneves , I believe we need to think the original proposed fix.

I've come up with another solution, which involves simply testing the build. I think this will be much more reliable. As before, tested on x86_64 and armv71. Some additional tests would be appreciated.

khovratovich:
I support the idea, please inform when you get all the tests.

technion:
Paging @daxtens  for a test on the updated strategy on your platform.

It looks like we can add OSX to the tested OK platforms based on reading the Travis log.

daxtens:
A quick eye-balling suggests that will work; I'll test it once I get into the office.

daxtens:
Works for me! :+1:

technion:
@khovratovich Positive test from @daxtens, and my colleagues that tested. Would be great to see this merged.

lucab:
I'm not an expert in cross-compilation, but I guess the `-march=native` would break in that scenario. Can we make this opt-out, so the auto-detection is skipped if `OPTTEST` is already explicitly set?

daxtens:
I'm not quite sure that I understand what you mean. If the target (be it native or cross compiled) doesn't support both `-march=native` and the SSE instructions in `opt.c`, the compilation will fail and `OPTTEST` will be set to a non-zero value, causing the reference implementation to be used.

ranisalt:
I think I understand what @lucab means. IIRC, under cross compilation, `-march=native` enables flags for the host machine, not the target, thus compiling with flags that may not be supported.

`-march=native` is not the kind of flag to be enabled on a Makefile but rather by environment variables supplied by the user. That said, instead of `-march=native`, only the needed flags should be enabled, with `-march` incrementing optimization.

daxtens:
Hmm. 

Running on an x86 box with a powerpc64le cross compiler:
```
$ powerpc64le-linux-gcc -march=native
powerpc64le-linux-gcc: error: unrecognized command line option ‘-march=native’
powerpc64le-linux-gcc: fatal error: no input files
compilation terminated.
```

I have never heard of cross-compilers working the way you suggest - do you have an example?

technion:
@ranisalt What you're missing is that no user in general is going to set such flags if they are just running "make" themselves. In most cases it's a trivial performance change, but Argon2 has a completely separate, optimised implementation that is meant to be used on x86 machines, which takes heavy advantage of this specific flag.

I note you hardcoded these into your JavaScript bindings - which means any non-x86 server won't be able to run them at all. With the rise of ARM, this all or nothing approach means those bindings are useless on an array of platforms. It's the same problem I have in Ruby and the problem I'm aiming to resolve here.

ranisalt:
No, that's not what I said, I'm saying that `-march=native` is not appropriate **only for cross compiling**, other uses are fine, so maybe it's not good to hardcode it, not that you did anything wrong. Actually I think this should be a separate issue and this PR is just fine.

daxtens:
Cool, let's get this merged, and then deal with whether it shouldn't be hard coded.

veorq:
Thanks guys for the testing

PR #116, opened 2016-03-08
Salt length is not at most 16 characters, but at least 8 characters

jedisct1:


ranisalt:
Good, missed my eye. There's also the README to fix.

veorq:
README fixed in 9623a7a2e0e1e6be6a20d736963339cc6c13e446

PR #119, opened 2016-03-17
Decode

khovratovich:
New output format is introduced with version number explicitly printed. Older outputs are treated with version number 0x10, for which the legacy code is added (with memory overwrite, not XOR).

@veorq @sneves Please check and approve


veorq:
Good here, though it's unfortunate to have to pass version number to argon2_hash, but I don't see any better solution.

include/argon2.h
```diff
@@ -178,6 +178,8 @@ typedef struct Argon2_Context {
     uint8_t *out;    /* output array */
     uint32_t outlen; /* digest length */
 
+    uint32_t version; /*version number*/
```
sneves: This `version` member would seem to make more sense below, along with all the other Argon2-related hashing settings, instead of mixed with the inputs/outputs. This has no effect on anything, I just found the positioning of this member variable odd.

include/argon2.h
```diff
@@ -204,6 +206,12 @@ typedef struct Argon2_Context {
 /* Argon2 primitive type */
 typedef enum Argon2_type { Argon2_d = 0, Argon2_i = 1 } argon2_type;
 
+/* Version of the algorithm */
+typedef enum Argon2_version {
```
sneves: Maybe have something like `ARGON2_VERSION_10 = 0x10, ARGON2_VERSION_13 = 0x13, ARGON2_LATEST_VERSION = ARGON2_VERSION_13`?

include/argon2.h
```diff
@@ -342,6 +351,19 @@ ARGON2_PUBLIC int argon2_verify_ctx(argon2_context *context, const char *hash,
...
+/**
+ * Returns the encoded hash length for the given input parameters
+ * @param t_cost  Number of iterations
+ * @param m_cost  Memory usage in kibibytes
+ * @param parallelism  Number of threads; used to compute lanes
+ * @param saltlen  Salt size in bytes
+ * @param hashlen  Hash size in bytes
+ * @return  The encoded hash length in bytes
+ */
+ARGON2_PUBLIC uint32_t argon2_encodedlen(uint32_t t_cost, uint32_t m_cost,
```
sneves: This should not return `uint32_t`, but instead `size_t`. Consider the edge case where `hashlen = ARGON2_MAX_OUTLEN`, and we're on a 64-bit architecture.

src/encoding.c
```diff
@@ -405,3 +410,16 @@ int encode_string(char *dst, size_t dst_len, argon2_context *ctx,
 #undef SX
 #undef SB
 }
+
+uint32_t b64len(uint32_t len) {
```
sneves: Same here: return `size_t`.

sneves:
Didn't have a lot of time to look at this, but I've just left a few nitpick comments on the commit of various things I noticed.

khovratovich:
Thank you Samuel, I will correct these. Do you think we need more time for review?

sneves:
I don't suppose it's worth holding this on my account. If no one else has any complaints, go ahead.

daniel-dinu:
I addressed the above comments. 
@veorq @sneves If everything is fine and there are no other issues, you can proceed with the merge.

PR #122, opened 2016-03-27
Remove test with m=20, as it takes too much RAM for many devices.

technion:
The m=20 test burns a GB of RAM. On an EC2 instance with only 512MB RAM, this test runs mostly in swap space, and runs for over 40 minutes. Obviously, this is undesirable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ OBJ = $(SRC:.c=.o)
 
 CFLAGS += -std=c89 -pthread -O3 -Wall -g -Iinclude -Isrc
 CI_CFLAGS := $(CFLAGS) -Werror=declaration-after-statement -D_FORTIFY_SOURCE=2 \
-				-Wextra -Wno-type-limits -Werror -coverage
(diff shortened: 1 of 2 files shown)


sneves:
I agree that this much memory should not be assumed to exist everywhere. At the same time, I don't quite like to remove the test altogether. Do you see any simple way to detect enough (physical) memory is available?

technion:
Certainly would agree that making it auto-detected would be a good idea, but I'm having an issue with portably identifying that in a portable manner. I've come across sample code involving inline ASM but would suggest that's a bad idea.

How would you feel if I configured it to be included only in the "testci" build - this would assert the high memory test was at least always tested in the CI environment.

sneves:
That seems pretty reasonable.

technion:
Updated commit shows the m=20 tests run in Travis, but not on my test VM when I run "make test". This ensures this large test is run upon commits, but doesn't bring limited end user machines to their knees. 

The remaining test suite runs in 12seconds on my VPS, which is a lot more tolerable than 40 minutes.

PR #129, opened 2016-04-07
Properly zero pwd pointed before using it.

technion:
argon2_verify creates an argon2_context, and configures a number of elements before passing it to decode_string. There, pwdlen is set to 0 and it is passed to validate_inputs.

There, pwd is used in a comparison as an uninitialized variable.


jedisct1:
Actually, both `.pwd` and `.pwdlen` should be initialized in the `_verify` function, as these are unrelated to what the `_decode` function decodes.

technion:
Good point @jedisct1, I'm pushing an updated commit.

PR #130, opened 2016-04-16
argon2: add manpage with usage info

lucab:
I'm forwarding this manpage which has been submitted to Debian bug tracker:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=820491

### DIFF
--- a/man/argon2.1
+++ b/man/argon2.1
@@ -1,0 +1,47 @@
+.TH ARGON2 "1" "April 2016" "argon2 " "User Commands"
+
+.SH NAME
+argon2 \- generate argon2 hashes
+
+.SH SYNOPSIS
+.B argon2 salt
+.RB [ OPTIONS ]
+
+.SH DESCRIPTION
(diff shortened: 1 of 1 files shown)


sneves:
Just a nitpick, there's a typo in line 17: "indepedent". I don't like very much the "this is the correct variant" wording, but I get that it's trying to get non-experts away from 2d.

lucab:
I've fixed the typo (which I didn't spot when proofreading) and reworded the last sentence to follow the text in the README (and avoid mentioning what's "correct").

Just as a note, the manpage is not going to be automatically installed (as there is no install target), but this should avoid downstream distributors to come up with multiple/different manpages.

Moreover, I'd like to move hash length option to `-l` and use `-h` for usage. I know this is bikeshedding, but is it ok to submit a PR for that?

sneves:
Thanks. Regarding the command line flag, it absolutely makes sense; I welcome that change.

PR #131, opened 2016-04-21
Ensure memset is not called with n == 0. This is undefined in C.

technion:
It turns out calling memset() with n == 0 is undefined behaviour. This is discussed here:

http://stackoverflow.com/questions/8597034/can-memset-be-called-with-a-null-pointer-if-the-size-is-0

[tis-interpreter](https://github.com/TrustInSoft/tis-interpreter) has identified that the codebase does currently make such a call.

I cannot find definitive information regarding the other functions, however:
- I presume memset_s behaves the same
- [this discussion](https://gcc.gnu.org/ml/gcc/2015-09/msg00135.html) seems to suggest explicit_bzero should check for n == 0
- There is no suggestion anywhere that SecureZeroMemory behaves this way, that I can find

The implementation is messier than I would like - because simply placing this check at the start of the function messes with C89 compliance, where the function pointer is declared afterwards.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -88,18 +88,12 @@ int argon2_hash(const uint32_t t_cost, const uint32_t m_cost,
     int result;
     uint8_t *out;
 
-    /* Detect and reject overflowing sizes */
-    /* TODO: This should probably be fixed in the function signature */
-    if (pwdlen > ARGON2_MAX_PWD_LENGTH) {
-        return ARGON2_PWD_TOO_LONG;
-    }
-
     if (hashlen > ARGON2_MAX_OUTLEN) {
         return ARGON2_OUTPUT_TOO_LONG;
     }
 
-    if (saltlen > ARGON2_MAX_SALT_LENGTH) {
-        return ARGON2_SALT_TOO_LONG;
+    if (hashlen < ARGON2_MIN_OUTLEN) {
+        return ARGON2_OUTPUT_TOO_SHORT;
     }
 
     out = malloc(hashlen);


sneves:
Can you explain how you got tis-interpreter to complain about this? As far as I understand, it's not the `n == 0` that is the problem, but `v == NULL`. Where is the latter happening?

technion:
Hi @sneves , to be more precise, quoting from @pascal-cuoq:

Please not that it is not calling  memset  or  memset_s  with a size of 0 that is invalid. It is calling them with an invalid pointer. In the past this was thought to be acceptable with a size of 0, but modern compilers apply optimizations that assume that the pointer arguments of these functions are valid.

To get it to complain, I created a threadless branch here (I really wish github would stop encouraging me to make a PR for this):

https://github.com/technion/phc-winner-argon2/tree/tis_stubs

And ran with run.sh.

sneves:
Thanks, it works. 

The problem here is that `argon2_hash`, called by `argon2_verify`, does not verify whether `hashlen < ARGON2_MIN_OUTLEN`. If `hashlen` is 0 (`decode_string` allows output to be empty, in which case `ctx->outlen = hashlen = 0`), `malloc` is allowed to return a non-`NULL` pointer that is nevertheless not writable and we get the UB report. 

This should have been rejected immediately by `validate_inputs`, but that only gets called in `argon2_ctx`, after we allocate memory.

Adding this check to `argon2_hash` makes tis-interpreter happy (or, maybe better, call `validate_inputs` after constructing `context`):

```
  if(hashlen < ARGON2_MIN_OUTLEN) {
    return ARGON2_OUTPUT_TOO_SHORT;
  }
```

technion:
Would the most correct solution be to implement both - with the secure_wipe_memory safe on its own, and a length check within argon2_hash to address this in the most appropriate place?

technion:
I suppose a related bikeshed here would be whether the pwdlen and saltlen checks in argon2_hash do anything, given the values then get passed to validate_inputs.

Perhaps I should replace that whole section with only a check for hashlen, against min and max values? I think this addresses the TODO in the most efficient way.

sneves:
I think that makes sense.

I'm on the fence about fixing `secure_wipe_memory` itself by checking for 0. I guess it's OK, though it might hide other semi-bugs hiding in other parts of the code that we haven't triggered yet (?).

technion:
Thanks for this commentary. I'm in favour of identifying future bugs, so I've removed the checks this PR originally introduced.

It now includes the hashlen checks, and has removed the superfluous checks tested in the validate_inputs() section.

tis-interpreter still produces a series of "inoccuous" warnings from unrelated code, which I'll address separately.

PR #132, opened 2016-04-27
Update README example code with newer context struct and function names.

technion:
The example in the README is out of date with regards to the argon2_context. I have also updated it to demonstrate safely checking return values, and added comments explaining fields.


technion:
I don't get what this code coverage tool is doing. This PR clearly only modifies a README. That means code coverage didn't change.

sneves:
Yeah, codecov looks confused.

SparkDustJoe:
Is it possible it didn't run on a previous merge or check-in?

PR #135, opened 2016-05-04
Removed some warnings about possible loss of data …

ola-ct:
… due to automatic conversions from `size_t` to `uint32_t`.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -212,7 +212,7 @@ int argon2_verify(const char *encoded, const void *pwd, const size_t pwdlen,
     }
 
     /* max values, to be updated in decode_string */
-    encoded_len = strlen(encoded);
+    encoded_len = (uint32_t)strlen(encoded);
     ctx.adlen = encoded_len;
     ctx.saltlen = encoded_len;
     ctx.outlen = encoded_len;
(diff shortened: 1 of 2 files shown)


technion:
What platform are you seeing these errors on? I've tested quite a few with all errors turned on, and have never seen any.

(just a question, and not a criticism of this PR).

ola-ct:
@technion The warnings occured when compiling with Visual Studio 2015 Community Edition (Debug/Release **x64**) on Windows 7 64 bit. They didn't occure when compiling to Debug/Release x86. 

With x64 being the destination platform it's obvious that data loss might occur because `strlen()` returns a `size_t` value which resolves to `unsigned __int64` which is double the size of `uint32_t` ;-)

lucab:
This looks like a real size_t/uint32_t type mismatch. Instead of casting/truncating, isn't it better to promote those variables to `size_t`, compare to `UINT32_MAX` and only then cast to `uint32_t` (or bail on errors)?

sneves:
I agree, the right thing to do here is to check against `UINT32_MAX`.

khovratovich:
I agree with @sneves . Can anyone make such a change?

PR #137, opened 2016-05-15
libargon2: add a pkg-config stub

lucab:
This is a pkg-config template to be used as a common base by downstream installers. Not used by us directly, as we don't provide any `install` target.


Leont:
I had completely missed this when writing #164, though I was already thinking we should add a pkg-config file. We should integrate these.

PR #146, opened 2016-07-01
Begin Separation

kmaragon:


kmaragon:
Trying to use cmdline utility to validate C# implementation and need to differentiate between lanes and threads because one very much matters to the hash and the other doesn't.

kmaragon:
oh whoops didn't mean to PR this way

PR #147, opened 2016-07-03
Add support for building DLLs with Visual Studio

PlasmaPower:
While previously Visual Studio would build the project just fine, the result would produce no exports, meaning that it would be useless as this is a library.


veorq:
thanks!

PR #148, opened 2016-07-22
Easier Windows DLL building + dependency-less builds

Qowyn:
Includes 2 new projects outputting DLLs and a new Configuration ReleaseStatic for creating binaries which only rely on kernel32.dll


khovratovich:
Thanks!

PR #149, opened 2016-07-25
Update languages to reflect binding implementation

ranisalt:
Now that a pretty lot of languages have bindings and some even have multiple styles, it would be nice to spot the differences fast. While some prefer native modules, some would prefer to deal with ffi, and I added that info to the language index.


ranisalt:
Wtf how come this will decrease coverage

veorq:
lol coverage decrease, we found a bug in codecov

ranisalt:
It's not the only PR with that problem hahahah

PR #153, opened 2016-08-20
Fix bad pointer copy when using custom allocator

WOnder93:
In the original code, the contents of the allocated memory are copied to instance->memory instead of the actual pointer value. This patch fixes the bug and also simplifies the code.


sneves:
Ouch. Thanks!

PR #154, opened 2016-08-20
fixed errors for mingw

gittermatt:
made changes to Makefile and src/thread.h to eliminate errors that where generated after "make" command using mingw

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ SRC_BENCH = src/bench.c
 SRC_GENKAT = src/genkat.c
 OBJ = $(SRC:.c=.o)
 
-CFLAGS += -std=c89 -pthread -O3 -Wall -g -Iinclude -Isrc
+CFLAGS += -std=c99 -pthread -O3 -Wall -g -Iinclude -Isrc
 CI_CFLAGS := $(CFLAGS) -Werror=declaration-after-statement -D_FORTIFY_SOURCE=2 \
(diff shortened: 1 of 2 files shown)


sneves:
What version of MinGW is this failing on?

gittermatt:
I know that it doesn't add support but mingw in my pc (version 4.8) generates error during the make presess that terminates building only argon2.exe and not the libraries, with my changes I'm able to build everything

gittermatt:
sorry, I previously said that the gcc version was 4.8, it is actually 4.9.3

sneves:
OK, I am able to reproduce the `CC=gcc` issue, but not the other ones. In particular, `uintptr_t` is already imported from the `process.h` definitions in every MinGW version I tried (which, granted, was not an exhaustive list).

Could you paste here the errors you were getting when building?

gittermatt:
error

gittermatt:
the error related to the c89 occours when building the test.c file

technion:
@gittermatt Am I correct in that the "test.c" you've referred to is your own test (possibly copied from the README) rather than distributed code? I can't find anything that looks like that line in the distribution's test.c.

I believe a more appropriate solution here would be to fix the example code, or build it specifically, rather than the whole project, using the relevant -std flag.

The CC= fix seems to be agreed on however.

gittermatt:
yes, test.c is the example in the readme and I agree with you, I just wanted to point out the error

khovratovich:
If the example is incorrect, there must be a change for readme.md, not for the thread.c and the makefile.

PR #155, opened 2016-08-21
makefile: optional explicit target for optimizations

lucab:
This commit allows passing an explicit OPTTARGET as the target
architecture for checking optimization capabilities.
This is needed for reataining compatibility to a baseline when
ditributing binaries.

For reference: https://bugs.debian.org/828779


lucab:
@veorq thanks for merging. We have now piled up quite a bit of commits/bugfixes since last tag, which I would like to have in Debian without massive backporting. Can you please tag a new one?

veorq:
There you go:
https://github.com/P-H-C/phc-winner-argon2/releases/tag/20160821

PR #156, opened 2016-08-29
add javascript browser port

antelle:
Hello!
I've compiled Argon2 for browsers with asm.js and WebAssembly, please add my port to ports list.


veorq:
Thanks!

PR #159, opened 2016-09-12
Refactor fill_segment code

WOnder93:
This PR attempts to de-duplicate and simplify the code for the fill_segment function and its helper functions.

Summary of changes:
- fill_block and fill_block_with_xor are merged into a single function
- fill_block in opt.c now accepts block pointers instead of uint8_t pointers
- Argon2i pseudo-random addresses are now computed on-the-fly instead of pre-computing them into a dynamically allocated buffer
- the above change also fixes a serious bug where fill_segment would silently give up and do nothing if the allocation failed

### DIFF
--- a/src/opt.c
+++ b/src/opt.c
@@ -21,13 +21,23 @@
 #include "blake2/blake2.h"
 #include "blake2/blamka-round-opt.h"
 
-void fill_block(__m128i *state, const uint8_t *ref_block, uint8_t *next_block) {
+void fill_block(__m128i *state, const block *ref_block, block *next_block,
+                int with_xor) {
     __m128i block_XY[ARGON2_OWORDS_IN_BLOCK];
-    uint32_t i;
+    unsigned int i;
 
-    for (i = 0; i < ARGON2_OWORDS_IN_BLOCK; i++) {
-        block_XY[i] = state[i] = _mm_xor_si128(
-            state[i], _mm_loadu_si128((__m128i const *)(&ref_block[16 * i])));
+    if (with_xor) {
+        for (i = 0; i < ARGON2_OWORDS_IN_BLOCK; i++) {
+            state[i] = _mm_xor_si128(
+                state[i], _mm_loadu_si128((const __m128i *)ref_block->v + i));
+            block_XY[i] = _mm_xor_si128(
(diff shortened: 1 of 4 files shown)


khovratovich:
It was intentional to have fill_block and fill_block_with_xor as separate functions for the clarity of the code.

WOnder93:
Well, I don't see how having two functions that do basically the same makes the code more clear. I would actually argue it does the opposite. Like this, if someone wants to read the code to see what the difference between these two functions is, they have to compare the functions line-by-line, to make sure they didn't overlook something. With a single function, the reader can just follow the if statements and see the difference much more easily.

WOnder93:
May I please request some more feedback on this? I can open a new PR without the merged fill_block functions if that is preferred, but I still believe that change should also be considered.

khovratovich:
If there is no performance drop, I approve.

josephlr:
Looks like there is no performance difference.

@WOnder93 version
```bash
echo "password" | ./argon2 "saltsalt" -t 1 -m 20 -p 1 -l 8
Type:   Argon2i
Iterations: 1 
Memory:   1048576 KiB
Parallelism:  1 
Hash:   837c82366605d74d
Encoded:  $argon2i$v=19$m=1048576,t=1,p=1$c2FsdHNhbHQ$g3yCNmYF100
0.796 seconds
Verification ok
```

Existing version
```bash
echo "password" | ./argon2 "saltsalt" -t 1 -m 20 -p 1 -l 8 
Type:   Argon2i
Iterations: 1 
Memory:   1048576 KiB
Parallelism:  1 
Hash:   837c82366605d74d
Encoded:  $argon2i$v=19$m=1048576,t=1,p=1$c2FsdHNhbHQ$g3yCNmYF100
0.797 seconds
Verification ok
```

WOnder93:
Yes, when I measured the speed of fill_block alone, my version seemed to be about 1-2% faster (thanks to removing the dynamic memory allocation), so the overall performance should be about the same (if not very slightly better).

veorq:
@khovratovich I don't expect performance issues from those changes, I'll let you merge if you approve

khovratovich:
Thank you! Merged in  fbcdbc7

PR #163, opened 2016-09-21
Eliminate redundant function.

technion:
Both burn() and secure_wipe_memory() serve identical functions.

This patch addresses code reuse. I also have the following concerns about the existing code:
- burn() does not have the NOT_OPTIMIZED attribute. 
- burn() does not OpenBSD or Windows specific variants

I'm unsure how important some of these are, but discrepancies like this are a good reason to reuse functions. I don't *think* it matters that a prototype doesn't use the attribute, since nothing is compiled there.


veorq:
No need for two functions indeed, thanks!

PR #164, opened 2016-09-22
Add install target

Leont:
This adds an install target to the makefile, making it easier to install or package Argon2


veorq:
Thanks!

PR #168, opened 2016-10-04
Adding in Argon2id functionality

josephlr:
As discussed previously, this pull request adds back in the Argon2id type. This involves:
 - Adding in a new argon2_type, and changing the hashing function to
   recognize this new type.
 - Changing the interface to add argon2id versions of all the functions.
 - Changing the commandline tool to allow for the use of the new type.
 - Changing the README to document the new type.
 - Adding tests to make sure the new type is working.

Tested change by setting parameters to the same as in
https://github.com/khovratovich/Argon2/blob/master/TestVectors/Argon2id.txt
and comparing the output. Output was the same.

Also, added license headers to the blake2 files


veorq:
@josephlr Some of the Travis builds fail, can you please have a look?

sneves:
It's due to some C99-style declarations, and Travis using `-Werror`. Easy fix. There are also some typos in the documentation, e.g., `combonation`.  Again, nothing major.

josephlr:
I think I fixed all the spelling errors and c89 compatibility issues. The test coverage is down because right now we do not test the various modes in test.c. We instead use the test vectors in kats/ 

Also, sorry for the numerous commits, just squash them when you merge this into master.

sneves:
Looks good to me.

veorq:
Nitpick: shouldn't "data-depending and data-independent" be "data-dependent ..." instead?

PR #169, opened 2016-10-10
Add argon2themax binding

jdconley:
Include my new library in the Bindings section.

To be fair, it is just a proxy to the argon2 lib by @ranisalt. However, argon2themax includes a bench-like utility which will automatically choose the most costly Argon2 options within a given clock time ceiling. This allows engineers to easily choose an option that is more secure than the default on whatever system they are running on and still maintain a good user experience.

For instance, you can ask for an option set that takes up to 100ms to run, and the library will march through increasingly costly options to find a set that hits as close to 100ms, without going over, as possible.

Since it isn't strictly a new binding maybe there should be another section in the README for stuff like this?


veorq:
Makes sense to add it, thanks!

PR #170, opened 2016-10-18
fix memory leak in fill_memory_blocks

yomgui1:
This patch fixes memory leak of a localy heap allocated 'threads' variable in fill_memory_blocks().
The leak happens in case of failure of argon2_thread_join(), needed to wait end of just created threads.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -277,6 +277,7 @@ int fill_memory_blocks(argon2_instance_t *instance) {
         for (s = 0; s < ARGON2_SYNC_POINTS; ++s) {
             int rc;
             uint32_t l;
+            int join_error = 0;
 
             /* 2. Calling threads */
             for (l = 0; l < instance->lanes; ++l) {
@@ -318,8 +319,11 @@ int fill_memory_blocks(argon2_instance_t *instance) {
                  ++l) {
                 rc = argon2_thread_join(thread[l]);
                 if (rc) {
-                    return ARGON2_THREAD_FAIL;
+                    join_error = 1;
                 }
+            }
+            if (join_error) {
+                return ARGON2_THREAD_FAIL;
             }
         }
 
--- a/src/core.c
+++ b/src/core.c
@@ -278,7 +278,6 @@ int fill_memory_blocks(argon2_instance_t *instance) {
 
     for (r = 0; r < instance->passes; ++r) {
         for (s = 0; s < ARGON2_SYNC_POINTS; ++s) {
-            int rc;
             uint32_t l;
 
             /* 2. Calling threads */
--- a/src/core.c
+++ b/src/core.c
@@ -317,8 +317,7 @@ int fill_memory_blocks(argon2_instance_t *instance) {
             /* 3. Joining remaining threads */
             for (l = instance->lanes - instance->threads; l < instance->lanes;
                  ++l) {
-                rc = argon2_thread_join(thread[l]);
-                if (rc) {
+                if (argon2_thread_join(thread[l])) {
                     rc = ARGON2_THREAD_FAIL;
                     goto fail;
                 }
--- a/src/core.c
+++ b/src/core.c
@@ -287,8 +287,7 @@ int fill_memory_blocks(argon2_instance_t *instance) {
 
                 /* 2.1 Join a thread if limit is exceeded */
                 if (l >= instance->threads) {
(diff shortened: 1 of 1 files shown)


yomgui1:
closing: missing a line and need tests to make Codecov happy

yomgui1:
previous commit is a fixup of the first one

yomgui1:
I don't know how to fix the issues with codecov, as it seems the project doesn't test return value of argon2_thread_join (If I understand well).
Someone to give hints?

josephlr:
One potential thing you could do here is pull the `rc` variable out of the loops, and implement a `goto fail` type cleanup, so we free in one place. This might be preferable to having the cleanup/return done in 6 different places (4 `ARGON2_THREAD_FAIL` and 2 `ARGON2_MEMORY_ALLOCATION_ERROR`) and having an awkward `join_error` variable.

Something like:
```c
{
  uint32_t r,s;
  argon2_thread_handle_t *thread = NULL;
  argon2_thread_data *thr_data = NULL;
  int rc = ARGON2_OK;

...
    rc = argon2_thread_join(thread[l]);
    if (rc != ARGON2_OK) {
        goto fail;
    }  

...

fail:
  if (thread != NULL) {
      free(thread);
  }
  if (thr_data != NULL) {
      free(thr_data);
  }
  return ret;
}
```

A similar pattern is in the `blake2b()` function in `src/blake2/blake2b.c`.

This should also make codecov happy (as it is a line based metric), and the above is the traditional way to do this sort of cleanup.

yomgui1:
@josephlr yeah, it seems to be a correct solution. I was not confident to touch so much of code just for that. But refactoring error handling over the whole function seems to be needed and good.
I'm going ahead and prepare a new patch.

yomgui1:
@josephlr just one thing: I need to keep the join_error variable as it's preferable to join all threads even if some join fails, before function exit.

yomgui1:
@josephlr finally I've removed the idea to change the join behaviour. Could be another non-related pull-request. I just focus on the memory leak here.

josephlr:
@yomgui1 Agreed. There are two issues here concerning failure when we still have unjoined threads.

The first is what you talked about regarding failure at join time. The other is failure at creation time. I make 3 threads, creating the 4 fails, and I don't cleanup the first three.

If we do that sort of thing, we would probably want some sort of unified way to cleanup. It also might not matter, because if a create/join has failed, we can't rely on other joins succeeding.

josephlr:
Also, you probably won't be able to make codecov happy because we don't currently induce a thread failure in the tests.

It would be much nicer if the argon2 thread functions returned the correct error codes, that way we don't use rc for both an argon2 error code and the 0 vs -1 value we get from the thread functions. If we don't change the thread functions, we should put the argon2 thread function calls inside the if statements.

Like:
```c
if (argon2_thread_join(thread[l])) {
    rc = ARGON2_THREAD_FAIL;
    goto fail;
}
```

Aside from that, in terms of stopping the leaks in a reasonable way this looks good.

sneves:
By "correct error codes", do you mean `ARGON2_THREAD_FAIL`?

yomgui1:
mostly ok now (except for bigger refactoring as notice upper and )
I recommand the commit squash before integration.

josephlr:
@yomgui1 Just one thing, the other two calls to the argon2 thread functions should be wrapped in the same way.

yomgui1:
@josephlr agreed, done

josephlr:
LGTM

josephlr:
@sneves Ya, I think it would make more sense for `argon2_thread_join()` and others to return `ARGON2_THREAD_FAIL`. Like how `allocate_memory()` in `src/core.c` returns `ARGON2_MEMORY_ALLOCATION_ERROR`.

yomgui1:
@sneves I agree for the unification on how and what values are used when error has to be reported. But it's could be a huge, in the sens of how many code line touched. It's preferable to open a new request for that.
BTW, thank a lot for all your help.

sneves:
By the way, a better way to implement this `fill_memory_blocks` function would be with memory barriers, and a single loop for creating and joining all the threads, where `goto fail` would include joining all the threads along with freeing the memory. 

The pthreads API has a native barrier implementation, which would make this simple. However, the Win32 API did not have this until Windows 8, so we'd have to implement this in the thread wrapper API.

Issue #143, opened 2016-06-17
inconsistency with argon2_verify() and the "associated data" field


eli-collins:
I was looking through the source to learn more about the hash string format, and noticed in the comment above `decode_string()` there's a reference to a "keyid" and "data" parameters, which seem to correspond to the `secret` and `ad` fields in the argon2_context. 

However, it looks like keyid isn't actually parsed by `decode_string()` or written by `encode_string()`; including that in the hash just results in a decode error.

On the other hand, the `data=` parameter *is* parsed by `decode_str()`, and written to `ctx.ad`.  However, `argon2_verify()` seems to silently discard this field, instead of passing along to `argon2_context()`.  

This results in the case where a hash of "password" such as `$argon2i$v=19$m=512,t=2,p=2$c29tZXNhbHQ$uEeXt1dxN1iFKGhklseW4w` will still verify correctly if a data parameter is inserted in the string, no matter what the data parameter is!  For example, `$argon2i$v=19$m=512,t=2,p=2,data=c29tZWRhdGE$c29tZXNhbHQ$uEeXt1dxN1iFKGhklseW4w` still verifies correctly against "password". 

I was expecting that second hash to either 1) fail to parse, or 2) have the `ctx.ad` value passed on to `argon2_context()`, in which case it would fail to verify because the digest no longer matched.  (In the case of option 2, I think the correct hash string for "password" would be `$argon2i$v=19$m=512,t=2,p=2,data=c29tZWRhdGE$c29tZXNhbHQ$KgHyCesFyyjkVkihZ5VNFw`).

I'm guessing the 'data' parameter was either a work-in-progress feature intended to be fleshed out, or a remnant of some old code that needs to be stripped.  But wanted to clarify, as I'm writing a Python wrapper, and wanted to make sure I accounted for edge cases. 

Thanks!

- Eli

sneves:
This looks like a legitimate bug. `argon2_verify` is calling `argon2_hash`, which does not have an `ad` input. The simplest solution seems to be adding `ad` and `adlen` to `argon2_hash`, though I don't quite like the gigantic number of arguments there.

`keyid` was meant to be an _identifier_ for a secret key; this effectively means that it can't be used in a generic library since it's application-dependent. The solution would be an input callback to transform an identifier into a key, but I'm not sure there's enough interest for that.

technion:
Is this something that is perhaps better addressed by removing keyid from the spec?

[...]

PR #173, opened 2016-10-25
Fixed up encoding issues

josephlr:
I significantly constrained the space of admissible encodings to deal with #143 . 

The `keyid` and `data` parameters in the encoding are gone, as they are either not used in validation, or they cause validation to fail. Similarly, the salt is now required. Before if no salt was provided, the encoding wouldn't include the salt or the hash value. Because there was no hash value, any later validation check would fail.

I also made changes to `validate_context`. Empty vs. `NULL` passwords and salts now produce the same results and errors. The salt was the real outlier here, a `NULL` salt was allowed while an empty salt produced an error (`ARGON2_SALT_TOO_SHORT`). This is inconsistent with the rest of Argon2, where switching a  parameter from empty to `NULL` produces the exact same hash value (or produces the exact same error).

Note that this doesn't break virtually any consumers of the public API in `argon2.h` because all encodings emitted from functions like `argon2_hash()` are still valid. There are only two cases where an existing consumer would see breakage:

1. They had manually added a `data` parameter to the string encoding. Note that this parameter has always been ignored.

2. They were hashing with a `NULL` salt (because hashing with an empty salt doesn't work). This is quite insecure, and they already would have been unable to use `argon2_verify()` as it would just fail when decoding the encoded string. 

I also added tests for a bad salt in an encoding string.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -354,37 +354,37 @@ int validate_inputs(const argon2_context *context) {
         return ARGON2_OUTPUT_TOO_LONG;
     }
 
-    /* Validate password length */
+    /* Validate password (required param) */
     if (NULL == context->pwd) {
         if (0 != context->pwdlen) {
             return ARGON2_PWD_PTR_MISMATCH;
         }
-    } else {
-        if (ARGON2_MIN_PWD_LENGTH > context->pwdlen) {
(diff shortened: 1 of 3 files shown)


veorq:
Thanks!

PR #180, opened 2016-11-03
Allow setting Argon2 version number from the command-line

Alexis-D:
Those flags allow one to make sure that the generated hashes are consistents over time even if the `argon2` binary is updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ results. `make install PREFIX=/usr` installs it to your system.
 on your system. To show usage instructions, run
 `./argon2 -h` as
 ```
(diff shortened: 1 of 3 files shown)


josephlr:
I like this idea, but to keep consistency with the other flags, maybe make the command line flags look like `-v 10` and `-v 13` instead of `-10` and `-13`. This would also require a check for an invalid version.

josephlr:
Nice! Do you think it is necessary to note that we default to using the most recent version if no version flag is specified?

Leont:
> but to keep consistency with the other flags, maybe make the command line flags look like -v 10 and -v 13 instead of -10 and -13

That sounds much more future-compatible to me

veorq:
Thanks!

PR #181, opened 2016-11-05
Add doc for argon2_encodedlen's type parameter

hynek:
Noticed this is missing while updating the Python bindings.


veorq:
Thanks!

PR #185, opened 2017-01-20
Shared library fix

### LINKED ISSUE #184: Run time failure for dynamically linked executable with default installation
## Steps to reproduce (on 64-bit Ubuntu Linux):

0. Have a clean system with no argon2 libraries installed
1. Clone Repository and run `sudo make install`.
2. Create a test file like this `test.c` example, which calls into argon2:
```c
// test.c
#include <stdio.h>
#include <argon2.h>

int main() {
  printf("%s\n", argon2_type2string(Argon2_id, 1));
  return 0;
}
```
3. Dynamically link in libargon2 by compiling the test against the argon2 shared library with `gcc -Wall test.c -largon2 -pthread -o test`.
4. While the compilation succeeds, running `./test` results in a run time failure: `./test: error while loading shared libraries: libargon2.so.0: cannot open shared object file: No such file or directory'.

The installer installs the shared object as `libargon2.so`, but it ends up being linked as `libaron2.so`. This can be seen by running `ldd test`:
```
joerichey@joerichey-ubiquity:~/tests/argon2-tests$ ldd test
	linux-vdso.so.1 =>  (0x00007ffe89120000)
	libargon2.so.0 => not found
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fcd48e43000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fcd49208000)
```

### Things not broken
We compiling the test as a _static_ executable, there are no problems.
```
joerichey@joerichey-ubiquity:~/tests/argon2-tests$ gcc -Wall -static test.c -largon2 -pthread -o test
joerichey@joerichey-ubiquity:~/tests/argon2-tests$ ./test
Argon2id
joerichey@joerichey-ubiquity:~/tests/argon2-tests$ ldd test
	not a dynamic executable
```
The linker/executable can find the static library `libargon2.a` but cannot find `libargon2.so`.

## Causes
This issues is caused by a bad `soname` of `libargon2.so.0` passed during the compilation of the shared library. [Here is the location in the Makefile](https://github.com/P-H-C/phc-winner-argon2/blob/master/Makefile#L54).

This `soname` argument means that while the linker will find the `/usr/lib/libargon2.so` file, it will dynamically link the executable to `/usr/lib/libaron2.so.0`, which then causes a failure at run time.

## Fixes
There are two ways to fix this problem. I have tested both methods on my system and they both resolve the issue.

### Fix 1 - Don't care about shared library versioning
For this fix, we just change the `soname` to be `libargon2.so` (or just remove that argument) and always install the shared library to `$PREFIX/lib/libargon2.so`. This means that a system can only have a single version of libargon2 on it at a time. This fix is simpler, but if libargon2 introduces ABI breaking changes and the dynamic library on a system is updated, old executables will break and will need to be relinked.

### Fix 2 - Deal with shared library versioning
For this fix, we keep track of our ABI version number (lets assume that number is 1). Then we install the shared object file to `$PREFIX/lib/libargon2.so.1` and have the `soname` also be `libargon2.so.1`. At installation time, we also symlink `$PREFIX/lib/libargon2.so` to `$PREFIX/lib/libargon2.so.1` (this symlink might not even be necessary). Then, when updating we can install the new version next to the old version and just update the symlink. 

This change is more complex, and we would now have to keep track of ABI breaking changes and bump the version number. If we went with this route, we would probably want to do something similar for the other platforms that support such a thing (like Darwin's `.dylib`). It is also important to note that the approach of the second fix is _significantly_ more common than that of the first.

josephlr:
The commit fixed #184, taking the "Fix 2" approach of dealing with the ABI Versioning.

This commit adds an `ABIVERSION` variable in the make file that should be incremented on every breaking change. Then, for Linux and macOS (Darwin), the name of the shared library is modified to include the version number. The unversioned shared library is then linked to this specific version.

As the install has gotten slightly more complicated, I also added an uninstall command for additional convenience. I also did a bit of formatting cleanup and set the correct permissions on the installed headers.

Tested on Ubuntu Linux (64 bit 3.13.0-105-generic) and macOS 10.11.6 (Darwin 15.6.0)

On linux:
Installed with `sudo make install`.
Compiled (with `gcc -Wall test.c -largon2 -o test`) and ran dynamically linked test from #184.
Compiled (with `gcc -Wall -static test.c -largon2 -pthread -o test`) and ran statically linked test from #184.
Uninstalled with `sudo make uninstall`

On macOS:
Ran Install/Uninstall with `make install PREFIX=~` and `sudo make install PREFIX=/usr/local`.
Compiled (with `gcc -Wall test.c -largon2 -o test`) and ran dynamically linked test from #184.
Compiled (with `gcc -Wall test.c /usr/local/lib/libargon2.a -o test`) and ran statically linked test from #184.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BENCH = bench
 GENKAT = genkat
 
 # Increment on an ABI breaking change
-ABIVERSION = 1
+ABI_VERSION = 0
 
 DIST = phc-winner-argon2
 
@@ -52,7 +52,7 @@ KERNEL_NAME := $(shell uname -s)
 
 LIB_NAME=argon2
 ifeq ($(KERNEL_NAME), Linux)
-	LIB_EXT := so.$(ABIVERSION)
+	LIB_EXT := so.$(ABI_VERSION)
 	LIB_CFLAGS := -shared -fPIC -fvisibility=hidden -DA2_VISCTL=1
 	SO_LDFLAGS := -Wl,-soname,lib$(LIB_NAME).$(LIB_EXT)
 	LINKED_LIB_EXT := so
@@ -62,7 +62,7 @@ ifeq ($(KERNEL_NAME), $(filter $(KERNEL_NAME),FreeBSD NetBSD OpenBSD))
 	LIB_CFLAGS := -shared -fPIC
 endif
 ifeq ($(KERNEL_NAME), Darwin)
-	LIB_EXT := $(ABIVERSION).dylib
+	LIB_EXT := $(ABI_VERSION).dylib
 	LIB_CFLAGS := -dynamiclib -install_name @rpath/lib$(LIB_NAME).$(LIB_EXT)
 	LINKED_LIB_EXT := dylib
 endif
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # Copyright 2015
 # Daniel Dinu, Dmitry Khovratovich, Jean-Philippe Aumasson, and Samuel Neves
 #
-# You may use this work under the terms of a Creative Commons CC0 1.0 
+# You may use this work under the terms of a Creative Commons CC0 1.0
 # License/Waiver or the Apache Public License 2.0, at your option. The terms of
 # these licenses can be found at:
 #
@@ -18,6 +18,9 @@
 RUN = argon2
 BENCH = bench
 GENKAT = genkat
+
+# Increment on an ABI breaking change
+ABIVERSION = 1
 
 DIST = phc-winner-argon2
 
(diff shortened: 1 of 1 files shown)


josephlr:
Questions for the maintainers:

1. Do you think we should also be using a minor version number for the shared libraries
2. Is the symlink to `libargon2.so` necessary. Some libraries do it, some do not.
3. Should the version variable be `ABIVERSION` or `ABI_VERSION`
4. What is the feasibility for adding support for the non-Windows UNIXs (BSD and SunOS)?

Leont:
Looks sensible enough to me.

khovratovich:
I can not answer these questions. Maybe @sneves @technion @lucab could comment?

technion:
I can't see this being a highly agile library in future, so I wouldn't think more than a single number would be needed.
Re: the symlink, most OS packages create this regardless of what libraries intend, so I'd assume it's a valid solution.

Makefile
```diff
@@ -19,6 +19,9 @@ RUN = argon2
 BENCH = bench
 GENKAT = genkat
 
+# Increment on an ABI breaking change
+ABIVERSION = 1
```
lucab: This is bumping the ABI version to 1 for no clear reason. Please keep in mind that every time this is bumped, all dependent packages need to be relinked.

lucab:
> Do you think we should also be using a minor version number for the shared libraries

No, I think that for the rate of change of this library a single number to bump is ok.

> Is the symlink to libargon2.so necessary. Some libraries do it, some do not.

It would be adviseable as it is useful for developers at link time, but not used at load time.

The latter is the main issue you are facing at https://github.com/P-H-C/phc-winner-argon2/issues/184: this library is already ABI-versioned (current version is `0`, resulting in soname `libargon2.so.0`) but the installer is not setting up files and paths in a proper way. You may want to have a look at the debian package to see how they should be arranged.

Makefile
josephlr: Agreed, the ABI version should start at 0.

josephlr:
@lucab I think leaving in the symlink is the right move, as there isn't a downside, and it is useful at link time.

khovratovich:
@josephlr Have you finalized the request?

sneves:
I agree that more than a single version number sounds like overkill for what this is. I can imagine bumping it if Argon3 showed up or something, but that would be a major version change either way. The PR looks good to me.

josephlr:
@khovratovich I don't have any more changes for the request. Looks good to merge in.

Issue #187, opened 2017-01-30
Useless thread creation for 1-thread version of argon2


luc-lynx:
According to the code at
https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.c#L310

if we need a 1-thread version of argon2 then that 1 thread is created by pthread library and then the main thread will wait for that one. It seems to be an overhead because for 1-thread version of argon2 we can call `fill_segment` function without creating separate thread for that. This improvement will make 1-thread version faster :)

khovratovich:
@sneves Could you comment?

sneves:
It's certainly wasteful to create a thread for the `p=1` case. The performance difference of not doing so won't be noticeable for typical cases, but it might affect the throughput if e.g. you're a server computing many `p=1` hashes simultaneously, in effect having 2x the number of threads as needed.

josephlr:
One thing to consider might be the scenario where you want to use argon2 but in a situation that dosn't have threading support. This sort of change could be extended to allow for argon2 to be compiled or used without threads (only allow p=1 of course).

luc-lynx:
@josephlr sounds sane. Actually, in this case we can allow p != 1 case by carrying out all computations sequentially (of cause it'll be slower) and add optional OpenMP support for that loop (the corresponding `#pragma` declarations will be ignored if there is no OpenMP support in the target system).

technion:
@josephlr Whilst I agree, that would take more than fixing an unneeded code path, and wouldn't get fixed unless someone spent time carefully inserting #ifdefs in the code. Not that I'm against it.

sneves:
It turns out to be a simple thing, I submitted a PR for it just now.

sneves:
I suppose this can be closed now.

PR #188, opened 2017-02-02
avoid thread creation in p=1 case

luc-lynx:
A pull request for issue #187

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -258,16 +258,27 @@ static void *fill_segment_thr(void *thread_data)
     return 0;
 }
 
-int fill_memory_blocks(argon2_instance_t *instance) {
+/* Single-threaded version for p=1 case */
+static int fill_memory_blocks_st(argon2_instance_t *instance){
+    uint32_t r, s, l;
+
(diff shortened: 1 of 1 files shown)


luc-lynx:
There is something wrong with the codecov because Travis builds were successful..

sneves:
Given how simple the main loop is without threads, i.e., 

```
    for (r = 0; r < instance->passes; ++r) {
        for (s = 0; s < ARGON2_SYNC_POINTS; ++s) {
            for (l = 0; l < instance->lanes; ++l) {
                argon2_position_t position = { r, l, (uint8_t)s, 0 };
                fill_segment(instance, position);
            }
        }
    }
    return ARGON2_OK;
```

would it not be cleaner to have one main loop for `threads = 1`, and a separate main loop for `threads > 1`, instead of having a single loop with the conditions inline?

SparkDustJoe:
For performance as well, the fewest branches and tests inside the loop the
better.

luc-lynx:
@sneves sure, thanks for the suggestion. I've updated the PR.

PR #189, opened 2017-02-06
add support for no threading

sneves:
As suggested by Joseph Richey in #187, this adds support for a situation where threads may not be available for one reason or another.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,15 @@ SRC_BENCH = src/bench.c
 SRC_GENKAT = src/genkat.c
 OBJ = $(SRC:.c=.o)
 
-CFLAGS += -std=c89 -pthread -O3 -Wall -g -Iinclude -Isrc
+CFLAGS += -std=c89 -O3 -Wall -g -Iinclude -Isrc
 CI_CFLAGS := $(CFLAGS) -Werror=declaration-after-statement -D_FORTIFY_SOURCE=2 \
 				-Wextra -Wno-type-limits -Werror -coverage -DTEST_LARGE_RAM
+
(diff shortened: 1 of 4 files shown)


khovratovich:
The BENCH setting uses multithreading too. Should not we modify the bench.c code accordingly?

sneves:
The only actual use of threads is in `core.c`. Everywhere else, including `bench.c`, sets the thread number in the parameters, but this will just reflect the number of lanes.

khovratovich:
Why do we have checks failed?

sneves:
Fixed. `CI_FLAGS` was using `CFLAGS` before it was entirely defined.

luc-lynx:
@sneves actually there is no need to leave only `p=1` case if `ARGON2_NO_THREADS` defined as for `p > 2` all the computations can be carried out sequentially in a loop in the `fill_memory_blocks_st` function. I can submit a separate PR if you agree.. :)

sneves:
Isn't this what already happens?

```
#if defined(ARGON2_NO_THREADS)
    return fill_memory_blocks_st(instance);
#else
    return instance->threads == 1 ? fill_memory_blocks_st(instance) : fill_memory_blocks_mt(instance);
#endif
```

When `ARGON2_NO_THREADS` is defined, any parallelism degree goes through `fill_memory_blocks_st`. Otherwise, if `ARGON2_NO_THREADS` is not defined it uses threads as necessary. Did you mean something else?

luc-lynx:
@sneves Oh, I see, you're right.. sorry for misleading you.

khovratovich:
Thanks!

PR #191, opened 2017-02-23
Two small fixes to the spec

mrijkeboer:
1. Fix allowed range of secret value K to match code.
2. Add type y for Argon2id.


khovratovich:
Thanks!

PR #192, opened 2017-02-23
Remove fill_block function from opt.h and ref.h.

mrijkeboer:
Remove the fill_block function from opt.h and ref.h since it is not used outside opt.c and ref.c.

### DIFF
--- a/src/opt.h
+++ b/src/opt.h
@@ -21,15 +21,4 @@
 #include "core.h"
 #include <emmintrin.h>
 
-/*
- * Function fills a new memory block and optionally XORs the old block over the new one.
- * Memory must be initialized.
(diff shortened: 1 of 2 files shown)


sneves:
Well, what this PR really shows is that `opt.h` and `ref.h` don't have any purpose. They can be removed entirely, and `fill_block` can be made `static` in the implementation files. The main function that `opt.c` and `ref.c` implement, `fill_segment`, is already declared in `core.h`.

Additionally, I think the `fill_block` comment should be moved to the place of definition instead of outright deleted. It is useful information.

mrijkeboer:
@sneves I agree with you and made a new pull request (#194).

PR #194, opened 2017-02-26
Delete opt.h and ref.h.

mrijkeboer:
This deletes the opt.h and ref.h headers as they are not used outside opt.c and ref.c. It also makes the fill_block functions static. This was suggested by Samuel Neves in a comment on pull request #192.


sneves:
Thanks!

PR #196, opened 2017-03-13
Fix use after free if threads > lanes

angelsl:
In [this `for` loop](https://github.com/P-H-C/phc-winner-argon2/blob/f15cf8bacd6594fdc51948959675bbf7d74666c3/src/core.c#L337), (reproduced below)

````c
uint32_t l;
for (l = instance->lanes - instance->threads; l < instance->lanes;
     ++l) {
    if (argon2_thread_join(thread[l])) {
        rc = ARGON2_THREAD_FAIL;
        goto fail;
    }
}
````

if `threads` is more than `lanes` (which is not specified to be illegal) `l` underflows. None of the threads will be joined and `argon2_ctx` may return before all the threads end, which means the `argon2_instance_t` instance is freed while threads are using it.

This mostly causes SIGSEGVs and SIGFPEs.

There doesn't seem to be any point if `threads` > `lanes` so just (silently) limiting `threads` to `lanes` fixes this.

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -72,6 +72,10 @@ int argon2_ctx(argon2_context *context, argon2_type type) {
     instance.threads = context->threads;
     instance.type = type;
 
+    if (instance.threads > instance.lanes) {
+        instance.threads = instance.lanes;
+    }
+
     /* 3. Initialization: Hashing inputs, allocating memory, filling first
      * blocks
      */

### CODE REVIEW

**Comment by angelsl** [comment]
Changing `l = instance->lanes - instance->threads` to `l = ((instance-> lanes < instance-> threads) ? 0 : (instance->lanes - instance->threads))` also fixes this.

Not sure which is better.

**Comment by josephlr** [comment]
I think silently reducing the number of used threads is fine. Basically, we're saying "you wanted us to use 5 threads but it only makes sense for us to use 3; it it won't affect the final hash".

We should probably add a test case in for this though.

**Comment by sneves** [comment]
What do you think of adding a check [here](https://github.com/P-H-C/phc-winner-argon2/blob/f15cf8bacd6594fdc51948959675bbf7d74666c3/src/core.c#L490) like this:

```
    /* No more threads than lanes */
    if ( context->threads > context->lanes ) {
        return ARGON2_THREADS_TOO_MANY;
    }
```

josephlr:
@sneves that would certainly be the shortest change, and the parameters that would now be failing were probably failing before.

angelsl:
@sneves

Do you want to put in a pull request for that?

sneves:
I'm happy with any of the fixes. Since nobody is complaining about this silent one, I guess I'll merge it as-is.

angelsl:
I tried to add a test but I couldn't get it to crash reliably.

Not sure if I should add it anyway.

sneves:
It consistently crashes here when using 1 lane and 4 threads.

PR #198, opened 2017-03-21
Fixed compiler warnings about potential dangerous casts

CodingMarkus:
Casting unsigned long to uint32_t is not guaranteed to be a safe operation as a long int can also be 64 bit wide and very often it is 64 bit wide. This change silences a couple of compiler warnings I got while building the code.


khovratovich:
I do not approve setting the default number of rounds to 16. This is inconsistent with the strategy we the designers currently promote. https://www.ietf.org/mail-archive/web/cfrg/current/msg09101.html

PR #206, opened 2017-04-23
Remove unused ALIGN macro

WOnder93:
The 'ALIGN' macro defined in core.h is not used anywhere, so it can be removed.

### DIFF
--- a/src/core.h
+++ b/src/core.h
@@ -20,14 +20,6 @@
 
 #include "argon2.h"
 
-#if defined(_MSC_VER)
-#define ALIGN(n) __declspec(align(16))
-#elif defined(__GNUC__) || defined(__clang)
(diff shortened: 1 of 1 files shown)


josephlr:
It looks like this used to be used to align the [block struct](https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.h#L58), so it would have a 128 bit alignment. This is because [the optimized implementation](https://github.com/P-H-C/phc-winner-argon2/blob/master/src/opt.c#L37-L72) would consider a block as an array of `__m128i`. But this was broken anyway because the [memory allocator](https://github.com/P-H-C/phc-winner-argon2/blob/master/src/core.c#L87-L111) does not give the proper 128-bit alignment. I say remove the macro and deal with the alignment issues later.

sneves:
Alright, merged.

PR #207, opened 2017-04-24
Allow null character in password for cli tool

ranisalt:
Creating test hashes for [node-argon2](https://github.com/ranisalt/node-argon2) I noticed the CLI tool does not have the ability to hash passwords containing null characters. As this is easily fixable by getting the password length from the call to `fread` instead of `strlen`, I made the change.

Additionally, it does not trim trailing newlines anymore.

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -197,10 +197,6 @@ int main(int argc, char *argv[]) {
         fatal("Provided password longer than supported in command line utility");
     }
 
-    if (pwd[pwdlen - 1] == '\n') {
-        --pwdlen;
-    }
-
     salt = argv[1];
 
     /* parse options */
--- a/src/run.c
+++ b/src/run.c
@@ -105,7 +105,7 @@ static void run(uint32_t outlen, char *pwd, size_t pwdlen, char *salt, uint32_t 
     }
 
     if (!salt) {
-        clear_internal_memory(pwd, strlen(pwd));
+        clear_internal_memory(pwd, pwdlen);
         fatal("salt missing");
     }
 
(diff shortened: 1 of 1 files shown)


sneves:
This is fine by me, but isn't it a little weird that we're still deleting trailing newlines from what is increasingly a 'raw' byte input?

josephlr:
The example in the `README` also uses `echo -n "password" | argon2` which creates the impression that newlines are not stripped, when in fact they are.

ranisalt:
That's actually fine by me and I didn't notice since it's another issue, but I commited it anyway.

src/run.c
```diff
@@ -109,7 +109,6 @@ static void run(uint32_t outlen, char *pwd, char *salt, uint32_t t_cost,
         fatal("salt missing");
     }
 
-    pwdlen = strlen(pwd);
```
josephlr: There are a bunch of other uses of `strlen(pwd)` in this function. Now that pwd isn't necessarily null terminated, those should be changed to `pwdlen`.

ranisalt: You are right. I wasn't catching it since when `strlen(pwd)` = 128 it didn't run the function, so there was always a `\0` at the end. Nevertheless I changed since the function call was unnecessary.

veorq:
Any objection to (not) merge?

What are the good reasons to support null characters?

sneves:
Looks fine to me.

ranisalt:
> What are the good reasons to support null characters?

`0` is a byte just like all other numbers between `1` and `255`, why disallow this one specifically? It's one less exception to reason about :)

PR #208, opened 2017-04-25
BLAMKA with AVX2 optimisations

luc-lynx:
This is the first version of AVX2 optimised BLAMKA function. Any feedback and suggestions are much appreciated.

As far as I see Blake2B version is not vectorised - it looks like reference version is used, so I can also add AVX2 etc. optimisations for it.


src/core.h
```diff
@@ -20,6 +20,14 @@
...
 #include "argon2.h"
 
+#if defined(_MSC_VER)
+#define ALIGN(n) __declspec(align(16))
+#elif defined(__GNUC__) || defined(__clang)
+#define ALIGN(x) __attribute__((__aligned__(x)))
+#else
+#define ALIGN(x)
+#endif
+
```
WOnder93: This macro has been removed in commit 7c57b3f08c8630c521916cad329c022cb38fca63. You should not reintroduce it, unless you need to use it.

luc-lynx:
@WOnder93 Thanks, I've removed that macroses. I apologise profusely for that.

src/blake2/blamka-round-opt.h
```diff
@@ -176,5 +177,152 @@ static BLAKE2_INLINE __m128i fBlaMka(__m128i x, __m128i y) {
                                                                                \
         UNDIAGONALIZE(A0, B0, C0, D0, A1, B1, C1, D1);                         \
     } while ((void)0, 0)
+#else /* __AVX2__ */
 
-#endif
+#include <immintrin.h>
+
+#define rotr32(x)   _mm256_permutevar8x32_epi32(x, _mm256_setr_epi32(1, 0, 3, 2, 5, 4, 7, 6))
```
WOnder93: You could also use `_mm256_shuffle_epi32(x, _MM_SHUFFLE(2, 3, 0, 1))` here. That instruction takes an 8-bit immediate instead of loading a 256-bit constant, so it might be faster. (I didn't measure it, though)

luc-lynx:
Thanks for the suggestion! I've updated the PR.

Qix-:
> macroses

Sounds like a deadly programming-related disease. (:

khovratovich:
Any benchmark results?

luc-lynx:
@khovratovich Please find them at [pastebin](https://pastebin.com/raw/NX35GLKB). SSE version is on the left and AVX version is on the right. I carried out the benchmark on an AWS `m4.2xlarge` instance.

khovratovich:
Thank you for the PR!

Issue #211, opened 2017-05-28
Fine-grained memory setting


goblin:
The argon2 CLI utility only allows setting memory via _n_, to 2^n KiB. However, if I'm reading it right, the paper says that the algorithm can use any multiple of 4p KiB of memory, where _p_ is parallelism:

> • Memory size m can be any integer number of kilobytes from 8p to 2^32 − 1. The actual number of blocks is m', which is m rounded down to the nearest multiple of 4p.

I'd like to use 11GiB of memory, why doesn't the CLI let me?

Thanks

sneves:
I'm not a fan of changing the meaning of `-m` at this point, but I suppose an argument of a fine grained amount of memory could be added.

PR #212, opened 2017-06-01
add support for fine-grained memory setting

goblin:
to fix https://github.com/P-H-C/phc-winner-argon2/issues/211

This is the simplest patch I could think of. Let me know whether/how I should add tests for it (i.e. should I change the way all existing tests are using m_cost, or add some new cases, or something else).

The mnemonic I have for `-k` is "KiBs of memory" - suggestions for a different letter are welcome.


goblin:
Sorry I'll open a new one, I forgot to update my local repo on this one.

PR #213, opened 2017-06-01
add support for fine-grained memory setting

goblin:
to fix https://github.com/P-H-C/phc-winner-argon2/issues/211

This is the simplest patch I could think of. Let me know whether/how I should add tests for it (i.e. should I change the way all existing tests are using m_cost, or add some new cases, or something else).

The mnemonic I have for `-k` is "KiBs of memory" - suggestions for a different letter are welcome.

### DIFF
--- a/src/run.c
+++ b/src/run.c
@@ -50,7 +50,7 @@ static void usage(const char *cmd) {
            T_COST_DEF);
     printf("\t-m N\t\tSets the memory usage of 2^N KiB (default %d)\n",
            LOG_M_COST_DEF);
-    printf("\t-k N\t\tSets the memory usage N KiB (default %d)\n",
+    printf("\t-k N\t\tSets the memory usage of N KiB (default %d)\n",
            1 << LOG_M_COST_DEF);
     printf("\t-p N\t\tSets parallelism to N threads (default %d)\n",
            THREADS_DEF);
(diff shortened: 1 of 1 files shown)


src/run.c
```diff
@@ -37,8 +37,8 @@
 #define UNUSED_PARAMETER(x) (void)(x)
 
 static void usage(const char *cmd) {
-    printf("Usage:  %s [-h] salt [-i|-d|-id] [-t iterations] [-m memory] "
-           "[-p parallelism] [-l hash length] [-e|-r] [-v (10|13)]\n",
+    printf("Usage:  %s [-h] salt [-i|-d|-id] [-t iterations] [-m log2(memory in KiB)] "
+           "[-k memory in KiB] [-p parallelism] [-l hash length] [-e|-r] [-v (10|13)]\n",
```
josephlr: Maybe do something like `[-m log2(memory in KiB) | -k memory in KiB]` to indicate that both flags should not be used at the same time. Also, should there be checking that both `-m` and `-k` are not used?

goblin:
@josephlr I've done both, good points.

goblin:
hang on, it doesn't seem to work ;-)

goblin:
ok, seems to work fine, I forgot you need to specify salt before other arguments :-)

khovratovich:
Thanks!

PR #224, opened 2017-09-09
Remove dependency of inttypes.h

### LINKED ISSUE #205: Missing header file on VS2015
It's been reported on ranisalt/node-argon2#78 that there's no `inttypes.h` on VS2015. There's a polyfill project [here](https://code.google.com/archive/p/msinttypes/) but it probably misses the architecture-driven optimizations, yet seems to suffice and is used in libgit2 and ffmpeg.

That's a rather [large change](https://github.com/ranisalt/phc-winner-argon2/commit/4564ccdb3a073b263e086c672b4279ebffc7ec02) but might work.

ranisalt:
Fixes #205 and removes some includes that weren't necessary. Since the only required macro from that file was `PRIu32` I copied it to the file that needed it.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -25,7 +25,6 @@
 #endif
 #define VC_GE_2005(version) (version >= 1400)
 
-#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/src/genkat.c
+++ b/src/genkat.c
@@ -115,7 +115,7 @@ void internal_kat(const argon2_instance_t *instance, uint32_t pass) {
                     : ARGON2_QWORDS_IN_BLOCK;
 
             for (j = 0; j < how_many_words; ++j)
-                printf("Block %.4u [%3u]: %016" PRIx64 "\n", i, j,
+                printf("Block %.4u [%3u]: %016llx\n", i, j,
                        instance->memory[i].v[j]);
         }
     }
--- a/src/genkat.c
+++ b/src/genkat.c
@@ -15,7 +15,6 @@
  * software. If not, they may be obtained at the above URLs.
  */
 
-#include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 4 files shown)


sneves:
This PR breaks `genkat.c`, which uses `PRix64`. May I suggest, instead of defining a platform-specific `I32u`, do it instead as 
```
printf("Iterations:\t%lu \n", (unsigned long)t_cost);
printf("Memory:\t\t%lu KiB\n", (unsigned long)m_cost);
printf("Parallelism:\t%lu \n", (unsigned long)lanes);
```
and
```
printf("Block %.4u [%3u]: %016llx\n", i, j, (unsigned long long)instance->memory[i].v[j]);
```
`long` and `long long` are defined to be at least 32 and 64 bits respectively on the standard, so this will result in no unwanted truncations anywhere.

ranisalt:
I am not sure how that would sort out on Windows, that platforms has some quirks. Also, there are differences between 32 and 64 bit platforms.

sneves:
It works out on Windows; `unsigned long` is required to be >= 32 bits in every architecture, and `unsigned long long` is required to be >= 64 bits in every architecture. This is guaranteed. In many Unixes, `unsigned long` will be 64 bits, but there is no information loss in this conversion.

If you mean the format strings, `%llu` is the biggest compatibility issue, since it was introduced after Visual Studio 2005 only; earlier versions required `I64u`. But I find this acceptable.

ranisalt:
Not true, `unsigned long` is 4 bytes wide [on some memory models](http://en.cppreference.com/w/cpp/language/types). Namely, it is 8 bytes only on Windows x64, and 4 bytes everywhere else.

sneves:
`unsigned long != unsigned long long`.

ranisalt:
Yeah, just saw that I didn't make sense right after posting. Should be ok now.

sneves:
There's [this](https://travis-ci.org/P-H-C/phc-winner-argon2/jobs/274476786#L444) pointless warning left, which is fixable by a `(unsigned long long)` cast; otherwise we're good to go.

khovratovich:
@ranisalt Can you fix that warning by modifying the cast if it makes sense?

ranisalt:
Sure, done. It won't be a problem since it's output only, won't mess things internally.

PR #228, opened 2017-09-13
Add AVX-512F optimizations

WOnder93:
This PR adds support for the AVX-512F instruction set.

Cc: @solardiz


solardiz:
I think this should be merged as-is (to keep the addition of AVX-512 separate from cleanups to previously committed code), but then as a separate commit/PR `blamka-round-opt.h` could be cleaned up to use the same kind of `#if ... #elif ... #else ... #endif` logic that `opt.c` will use after this PR is merged. IOW, I suggest that we avoid the deep nesting of `#if`'s and consistently include definitions or code for the "higher" SIMD instruction sets first, followed by the "lower".

Leont:
> I think this should be merged as-is (to keep the addition of AVX-512 separate from cleanups to previously committed code), but then as a separate commit/PR blamka-round-opt.h could be cleaned up to use the same kind of #if ... #elif ... #else ... #endif logic that opt.c will use after this PR is merged. IOW, I suggest that we avoid the deep nesting of #if's and consistently include definitions or code for the "higher" SIMD instruction sets first, followed by the "lower".

That sounds sane to me

PR #229, opened 2017-09-27
saltlen change to size_t in run.c

### LINKED ISSUE #201: size_t and uint32_t confusion
In many places size_t and uint32_t types are used for the same things.

For example saltlen is size_t in argon*_hash* functions and uint32_t in Argon2_Context and argon2_encodedlen.

This makes my compiler to spit warnings, because on my machine size_t is 64 bytes long.

khovratovich:
argon2_encodedlen now takes saltlen of type size_t

Fixes #201

### DIFF
--- a/src/argon2.c
+++ b/src/argon2.c
@@ -445,8 +445,8 @@ const char *argon2_error_message(int error_code) {
 }
 
 size_t argon2_encodedlen(uint32_t t_cost, uint32_t m_cost, uint32_t parallelism,
-                         uint32_t saltlen, uint32_t hashlen, argon2_type type) {
+                         size_t saltlen, uint32_t hashlen, argon2_type type) {
   return strlen("$$v=$m=,t=,p=$$") + strlen(argon2_type2string(type, 0)) +
          numlen(t_cost) + numlen(m_cost) + numlen(parallelism) +
-         b64len(saltlen) + b64len(hashlen) + numlen(ARGON2_VERSION_NUMBER) + 1;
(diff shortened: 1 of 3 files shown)


src/argon2.c
```diff
@@ -445,8 +445,8 @@ const char *argon2_error_message(int error_code) {
 }
 
 size_t argon2_encodedlen(uint32_t t_cost, uint32_t m_cost, uint32_t parallelism,
-                         uint32_t saltlen, uint32_t hashlen, argon2_type type) {
+                         size_t saltlen, uint32_t hashlen, argon2_type type) {
   return strlen("$$v=$m=,t=,p=$$") + strlen(argon2_type2string(type, 0)) +
          numlen(t_cost) + numlen(m_cost) + numlen(parallelism) +
-         b64len(saltlen) + b64len(hashlen) + numlen(ARGON2_VERSION_NUMBER) + 1;
```
daniel-dinu: Replacing ``b64len(saltlen)`` with ``saltlen`` is not good!
See ``size_t b64len(uint32_t len)`` 
https://github.com/P-H-C/phc-winner-argon2/blob/d3d81d3002c65e3a1738dd11fdb19bf8ba23349d/src/encoding.c#L440-L453

I think a clean solution might be to do as suggested by in  https://github.com/P-H-C/phc-winner-argon2/issues/201#issuecomment-328027862

@khovratovich, @Leont: would it be possible to check the second option and comment or create a PR?

khovratovich: Could you provide a solution that works as you suggested?

PR #245, opened 2018-03-01
Allow changing PREFIX

AsavarTzeth:
Without this change it is not possible to change PREFIX to something
else, since the Makefile simply overrides it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,9 @@ HEADERS = include/argon2.h
 INSTALL = install
 
 DESTDIR =
+ifndef PREFIX
 PREFIX = /usr
+endif
 INCLUDE_REL = include
 LIBRARY_REL = lib
 BINARY_REL = bin


Leont:
> Without this change it is not possible to change PREFIX to something
> else, since the Makefile simply overrides it.

What do you mean with "not possible"? Doesn't `make PREFIX=whatever` work for you?

AsavarTzeth:
Sorry, I realise my description was a bit misleading. The answer is, well yes and no.

It works but only like you describe, but that means I also need to use `make install PREFIX=/app`. Initially this did not occur to me.

Instead, this simple change allows me to set PREFIX as environment variable, making this sufficient:
```
make
make install
```

This might seem like a minor thing, but it simplifies things a lot in build environments where a custom PREFIX would likely be set globally.

AsavarTzeth:
I see, I should know that already. In that case, I suppose this has been nothing more than a good reminder to me. I certainly would not want to trade security for a minor convenience.

Thank you for your time patience.

PR #251, opened 2018-04-06
Added Argon2 Erlang binding.

ergenius:
A few days ago i started building an Argon2 Erlang binding called eArgon2.

It's in his early beta stages, however it's working and bypassed simple tests. I added my binding to the binding list alphabetically (i noticed all are listed alphabetically). Please accept my listing. Thank's!


daniel-dinu:
Thanks!

PR #256, opened 2018-06-21
Added new Rust binding to README.md

bcmyers:
I recently created a new Rust library that wraps the C implementation of argon2: [argonautica](https://docs.rs/argonautica/0.1.2/argonautica/). It will also have a pure Rust implementation (as an additional option) in the future, but this is not yet written. Perhaps you could include it in the readme?


daniel-dinu:
Thanks!

PR #259, opened 2018-07-20
Remove the need for stdio.h (sprintf)

skeeto:
`stdio.h` is only needed for one `sprintf()`, which is only needed to trivially print a non-negative integer. This tiny change keeps all the printf machinery out of static builds.

### DIFF
--- a/src/encoding.c
+++ b/src/encoding.c
@@ -15,7 +15,6 @@
  * software. If not, they may be obtained at the above URLs.
  */
 
-#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


josephlr:
This change does seem to reduce the size of all the major targets (very slightly). It's not clear that this size reduction is worth the added complexity in the encoding code.

| Target | Size (master) | Size (this PR) | Change  |
|:--:|--:|--:|:--:|
| `libargon2.a` | `354.928 KB` | `351.544 KB` | `-0.95 %` |
| `libargon2.so` | `190.032 KB` | `188.792 KB` | `-0.65 %` |
| `argon2` (dynamic) | `252.736 KB` | `251.512 KB` | `-0.48 %` |
| `argon2` (static) | `1.614624 MB` | `1.613008 MB` | `-0.10 %` |

All sizes in bytes, built with optimizations for x86-64 on Ubuntu 18.04.

sneves:
I definitely like getting rid of `stdio.h` just for this `sprintf`, but the way this is done will lead to wondering what this loop does down the line.

Perhaps a static function documenting what the function does would be more maintainable?

PR #260, opened 2018-07-21
Fix typos

rffontenelle:
These were detected by running `codespell -ef`.


sneves:
I would merge this if it would remove the changes to `latex/IEEEtran.cls`, which is an external standard LaTeX template file.

rffontenelle:
Removed `latex/IEEEtran.cls` and git-squashed for a cleaner history.

sneves:
Thanks!

PR #261, opened 2018-08-31
Add tests for Argon2_id.

technion:
Hi,

With Argon2_id gaining popularity, it should be included in the tests.


khovratovich:
Thanks!

PR #264, opened 2019-01-08
argon2 executable: Add option -b which allows to provide the salt in base64 encoding

### LINKED ISSUE #263: Argon2 command line utility (run.c): Add possibility to encode Salt
Since currently the salt is directly interpreted by the command line utility it's not possible to use specific salts which contain for example 0-bytes, and other characters can be hard to transfer depending on the used shell.

A possible solution would be to introduce an additional command line parameter (-b). If this parameter is set, the salt is interpreted as base64.

charno:
Fixes #263.

Rationale:
A Salt in argon2 can be any binary set. Since the command line utility accepts the salt as argument, not all possible characters (for example null-byte) can be passed as salt. With the possibility to specify that the salt is transferred as base64, any binary set can be passed.

Implementation:
- use existing from_base64 function (remove static modifier)
- parameter handling the same as for other parameters

### DIFF
--- a/src/encoding.c
+++ b/src/encoding.c
@@ -145,20 +145,7 @@ static size_t to_base64(char *dst, size_t dst_len, const void *src,
     return olen;
 }
 
-/*
- * Decode Base64 chars into bytes. The '*dst_len' value must initially
- * contain the length of the output buffer '*dst'; when the decoding
- * ends, the actual number of decoded bytes is written back in
- * '*dst_len'.
- *
- * Decoding stops when a non-Base64 character is encountered, or when
- * the output buffer capacity is exceeded. If an error occurred (output
- * buffer is too small, invalid last characters leading to unprocessed
- * buffered bits), then NULL is returned; otherwise, the returned value
- * points to the first non-Base64 character in the source stream, which
- * may be the terminating zero.
- */
-static const char *from_base64(void *dst, size_t *dst_len, const char *src) {
+const char *from_base64(void *dst, size_t *dst_len, const char *src) {
     size_t len;
     unsigned char *buf;
     unsigned acc, acc_len;
--- a/src/encoding.h
+++ b/src/encoding.h
@@ -54,4 +54,19 @@ size_t b64len(uint32_t len);
 /* Returns the length of the encoded number num */
 size_t numlen(uint32_t num);
 
+/*
+ * Decode Base64 chars into bytes. The '*dst_len' value must initially
+ * contain the length of the output buffer '*dst'; when the decoding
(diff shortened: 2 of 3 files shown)


khovratovich:
I do not think we need to change the `run` signature for this.

charno:
> I do not think we need to change the `run` signature for this.

I did this since the other arguments are passed the same way. In the end, a boolean (to check if the salt has to be decoded) has to be present somehow in the `run` function.

D-Nice:
Something something unix philosophy?

```bash
argon2 $(echo dW5peGZ0dyEK | base64 -d)
```

I think should be available to vast majority of users, and realistically there's so many other utilities which encode/decode base64 perfectly fine. What if I want base58, or base58check? Security oriented applications should avoid bloat imho.

charno:
Hi,

Some Background: We have a C++ application which uses the argon2 library (as you suggested). This application creates and downloads argon2 hashed passwords and salts and stores them in a dtabase. Since the salts are generated as random bytearrays, they can contain 0-bytes.

One use-case is that we have a separate command line script which needs to check the password hashes which are stored in the database. That's where we run into the problems. For us it's not a solution to change the salt-format or to rehash the existing passowrds (since we don't know them).

I'm all for unix philosophy! And I did actually try something similar to your suggestion, but sadly it does not work for 0-bytes.

When I try with this command:
`echo asdf | argon2 $(echo MTIAMzQ1Njc4OQ== | base64 -d)`
(the base64-encoded string is 12\03456789, so contains a 0-byte), then I get different results depending on the shell used.
- ZSH: argon2-binary returns "Error: unknown argument" (possibly the hash gets split into two arguments?)
- bash: gives a warning that NULL byte in the argument is ignored
- busybox: simply ignores the NULL byte
- dash: simply ignores the NULL byte

With this pull request applied, the hashing works:
`echo asdf | argon2 MTIAMzQ1Njc4OQ== -b`, result: aec4e9d91d7315ca32260a514b646f3f62aa2f9c4e55c2365b50cba39244f529. 
The hash with the ignored 0-byte is 3a396fc5856b2b1073c5de7f439eae9afb80af9c5ad35f3bd5ee9e96cf576f46.

Personally, I don't care if the encoding of the salt would be done as base64 or hex or whatever, I just implemented base64 since we already have the salts as base64, and there was already a base64_decode function in the source which I could reuse. If you prefer any other format I can convert between formats in the shell just fine.

DragonWork:
+1 merge please

khovratovich:
@sneves @veorq What do you think of changing the `run` signature?

PR #266, opened 2019-02-10
Visual Studio improvements

noloader:
There are two improvements here. First, drop the minimum requirement for Visual  Studio to 2010 (Toolset version 10.00). This allows someone to open and run the project with VS2010, 201 2, 2013, 2015, 2017, 2019, etc. Whatever they have on their machine works, and they won't encounter "wrong toolset" errors (or similar). Second, enable debugging in formation so the code can be stepped under Visual Studio.

-----

This change came about due to testing on Windows 8.1 with Visual Studio 2010, 2012 and 2013 installed. The solution failed to compile because I lacked Visual Studio 2015 (Toolset version 14.00). I have Visual Studio 2015, 2017 and 2019 on a Windows 10 machine, but I despise working on Windows 10. It was better for me to fix the project files in a way so they "just work" everywhere.

Crypto++ has been using the `DefaultPlatformToolset` trick for years to avoid upgrades and broken compiles. Argon2 will be fine with the change.

### DIFF
--- a/Argon2.sln
+++ b/Argon2.sln
@@ -1,8 +1,6 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
-MinimumVisualStudioVersion = 10.0.40219.1
+Microsoft Visual Studio Solution File, Format Version 11.00
+# Visual Studio 2010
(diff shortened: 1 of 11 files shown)


noloader:
This change came about due to testing on Windows 8.1 with Visual Studio 2010, 2012 and 2013 installed. The solution failed to compile because I lacked Visual Studio 2015 (Toolset version 14.00). I have Visual Studio 2015, 2017 and 2019 on a Windows 10 machine, but I despise working on Windows 10. It was better for me to fix the project files in a way so they "just work" everywhere.

Crypto++ has been using the `DefaultPlatformToolset` trick for years to avoid upgrades and broken compiles. Argon2 will be fine with the change.

sneves:
Thanks!

PR #267, opened 2019-03-05
use explicit_bzero() on recent glibc versions

maciejsszmigiero:
glibc 2.25+ has explicit_bzero(), so we can use it to securely wipe memory
instead of hacking our own memset-based replacement, just like we already
do on OpenBSD.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -25,6 +25,9 @@
 #endif
 #define VC_GE_2005(version) (version >= 1400)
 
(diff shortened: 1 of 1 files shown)


sneves:
Thanks!

jedisct1:
FreeBSD >= 11.0 also have `explicit_bzero()`.

rashworld-max (approved):

PR #268, opened 2019-03-11
Wait for already running threads if a thread creation failed.

mbroz:
On memory-constrained systems (like cgroups limited processes)
thread creation often fails.

The code needs to wait for already running threads on error path;
otherwise these threads can access deallocated memory
(and cause a segfault or another crash).


sneves:
Yikes, nice catch!

rashworld-max (approved):

Issue #269, opened 2019-04-18
Install manual page


paulmenzel:
`make install` should also install the manual page.

    install -Dm644 man/argon2.1 ${DESTDIR}${MANDIR}/man1/argon2.1

Begasus:
Have been able to add this to the Makefile, will see to file a PR for that later :)

Begasus:
This is what I used to get the manpage installed through the Makefile (if there are any comments on it I'd like to hear it :) ) :
```
diff --git a/Makefile b/Makefile
index 1f24b06..5e61408 100644
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ RUN = argon2
 BENCH = bench
 GENKAT = genkat
 ARGON2_VERSION ?= ZERO
+MAN = man/argon2.1
 
 # installation parameters for staging area and final installation path
 # Note; if Linux and not Debian/Ubuntu version also add lib override to make command-line
@@ -176,6 +177,7 @@ INST_INCLUDE = $(DESTDIR)$(PREFIX)/$(INCLUDE_REL)
 INST_LIBRARY = $(DESTDIR)$(PREFIX)/$(LIBRARY_REL)
 INST_BINARY = $(DESTDIR)$(PREFIX)/$(BINARY_REL)
 INST_PKGCONFIG = $(DESTDIR)$(PREFIX)/$(PKGCONFIG_REL)/pkgconfig
+INST_MAN = $(DESTDIR)$(PREFIX)/$(MAN_REL)
 
 # main target
 .PHONY: all
@@ -259,6 +261,8 @@ endif
 	$(INSTALL) $(RUN) $(INST_BINARY)
 	$(INSTALL) -d $(INST_PKGCONFIG)
 	$(INSTALL) -m 0644 $(PC_NAME) $(INST_PKGCONFIG)
+	$(INSTALL) -d $(INST_MAN)
+	$(INSTALL) -m 0644 $(MAN) $(INST_MAN)
 
 .PHONY: uninstall
 uninstall:
-- 
2.21.0
```

Leont:
> This is what I used to get the manpage installed through the Makefile (if there are any comments on it I'd like to hear it :) ) :

I think that's missing a `MAN_REL?= man/man1`

Begasus:
> > This is what I used to get the manpage installed through the Makefile (if there are any comments on it I'd like to hear it :) ) :
> 
> I think that's missing a `MAN_REL ?= man/man1`

[...]

PR #273, opened 2019-06-15
install manpage

Begasus:


Begasus:
Think this should do it to install the manpage, not tested on other platforms beside Haiku, so could use some testers :) (PS, will create a separate PR to include Haiku in the platforms)

Begasus:
Related to https://github.com/P-H-C/phc-winner-argon2/issues/269

Begasus:
With a full patch to add Haiku specific paths it looks like this :) https://ibb.co/jySCFK7

Leont:
LGTM

Makefile
```diff
@@ -136,6 +137,7 @@ ifeq ($(KERNEL_NAME), $(filter $(KERNEL_NAME),DragonFly FreeBSD))
 BINARY_REL ?= bin
 INCLUDE_REL ?= include
 LIBRARY_REL ?= lib
+MAN_REL ?= man/man1
```
ryandesign, line 140: That should be share/man/man1.
```suggestion
MAN_REL ?= share/man/man1
```

Begasus: Sorry for the late reply, forgotten here :)
Pushed the changes, I'm think the share/man/man1 should be used in the other places also, also pushed the change to add Haiku specific paths

PR #274, opened 2019-06-15
Use autotools to automate build and pkg-config variable substituion

hughmcmaster:
This patch automates the build process using most of the existing handcrafted Makefile.
It's now the familiar process of running:
```
./configure --prefix=/usr
make
make install
```
Using autotools allows the dynamic variables in `libargon2.pc` to be automatically substituted using detected variables. It's much easier than the current manual process.

More could be done to further integrate autotools with the current buidl system, but this patch keeps most the existing build system to keep the changes small.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -67,47 +67,39 @@ MACHINE_NAME := $(shell uname -m)
 
 LIB_NAME = argon2
 PC_NAME = lib$(LIB_NAME).pc
-PC_SRC = $(PC_NAME).in
 
 ifeq ($(KERNEL_NAME), Linux)
 	LIB_EXT := so.$(ABI_VERSION)
 	LIB_CFLAGS := -shared -fPIC -fvisibility=hidden -DA2_VISCTL=1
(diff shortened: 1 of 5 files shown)


Leont:
There's already a PR to add CMake and Mason support, I'm not sure what this would add to that.

jedisct1:
Can we have xmake, GYP, waf, scons, makeme, premake, buck, tup, and build2, too? Thanks!

hughmcmaster:
> 
> 
> There's already a PR to add CMake and Mason support, I'm not sure what this would add to that.

TBH, I didn't look at the existing PRs before working on this. But I see no reason to add both CMake and Meson support -- one is sufficient. My intention with this patch was to keep the change small, since the project already used a handcrafted Makefile.

Also, that PR was opened in October last year, with no activity until recently.

hughmcmaster:
@jedisct1 Sarcasm doesn't make for a productive project.

hughmcmaster:
Hi there, any further thoughts on this PR?

PR #276, opened 2019-08-22
Adding new Android binding to README.md

lambdapioneer:
I've published an Android binding named [Argon2Kt](https://github.com/lambdapioneer/argon2kt) for Argon2 that makes it simple for Android developers to use Argon2. It emerged from a personal mobile app project.

Looking at the existing bindings I've found that the existing Java bindings don't work too well with Android - especially with the now mandatory 64-bit support. Argon2Kt can be used from both Kotlin and Java projects.

I hope adding to the listing will help other mobile developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ Argon2i 1 iterations  4096 MiB 4 threads:  2.72 cpb 11124.86 Mcycles
 Bindings are available for the following languages (make sure to read
 their documentation):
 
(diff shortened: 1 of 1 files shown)


lambdapioneer:
Friendly ping as I just released the v1.0 as the first stable version :)

khovratovich:
Thanks

PR #285, opened 2020-04-10
mingw-w64 / MSYS2 fixes

dra27:
These are drawn from patches in https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-argon2, with corrections.

`_MSC_VER` is defined specifically by the Microsoft C compiler and should be used for code _specific to that compiler, not Windows_. `_WIN32` is defined by any C compiler targeting Windows *automatically* - i.e. both the Microsoft C compiler and mingw-w64 GCC define it.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -16,7 +16,7 @@
  */
 
 /*For memory wiping*/
-#ifdef _MSC_VER
+#ifdef _WIN32
 #include <windows.h>
 #include <winbase.h> /* For SecureZeroMemory */
 #endif
@@ -132,7 +132,7 @@ void free_memory(const argon2_context *context, uint8_t *memory,
 #endif
 
 void NOT_OPTIMIZED secure_wipe_memory(void *v, size_t n) {
-#if defined(_MSC_VER) && VC_GE_2005(_MSC_VER)
+#if defined(_MSC_VER) && VC_GE_2005(_MSC_VER) || defined(__MINGW32__)
     SecureZeroMemory(v, n);
 #elif defined memset_s
     memset_s(v, n, 0, n);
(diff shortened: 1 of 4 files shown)


src/core.c
```diff
@@ -132,7 +132,7 @@ void free_memory(const argon2_context *context, uint8_t *memory,
 #endif
 
 void NOT_OPTIMIZED secure_wipe_memory(void *v, size_t n) {
-#if defined(_MSC_VER) && VC_GE_2005(_MSC_VER)
+#if defined(_MSC_VER) && VC_GE_2005(_MSC_VER) || defined(__MINGW32__)
```
dra27, line 135: I'm not certain why this has been tagged as Visual Studio 2005+, so I've used the `__MINGW__`, rather than changing it to `_WIN32` (NB the 64-bit version of mingw-w64 also defines `__MINGW32__`).

src/genkat.c
```diff
@@ -115,7 +121,7 @@ void internal_kat(const argon2_instance_t *instance, uint32_t pass) {
                     : ARGON2_QWORDS_IN_BLOCK;
 
             for (j = 0; j < how_many_words; ++j)
-                printf("Block %.4u [%3u]: %016llx\n", i, j,
+                printf("Block %.4u [%3u]: %016" PRIx64 "\n", i, j,
```
WOnder93, line 124: Does this file really need changing? The value is already being cast to `unsigned long long`, so `%llx` should be appropriate. Or did you intend to remove the cast below?

dra27: I don’t follow - the issue on mingw is that `printf` doesn’t always understand `%llx` and the definition will substitute the appropriate Microsoft `%I64` version if needed.

WOnder93: Ah, I see. I didn't expect there are compilers that don't understand the `llx` format...

And I guess since you fallback to `"llx"` on non-mingw, then you can't safely drop the cast to `unsigned long long`... Never mind my comment then.

PR #286, opened 2020-04-10
Allow compilation of mingw-w64 from Cygwin cross-compilers

dra27:
This merges in patches from downstream at https://github.com/cygwinports/argon2 to conform with Cygwin DLL naming conventions. At a high level changes, the changes mean:

- Cygwin builds `cygargon2-1.dll` with `libargon2.dll.a` as the import library
- "mingw" (this doesn't really exist any more) also uses the `$ABI_VERSION` but keeps the more normal `lib` prefix so builds `libargon2-1.dll` with `libargon2.dll.a`. This corresponds with `sys-root` installations of mingw-w64 libraries on Cygwin
- MSYS2 is untouched, since downstream makes no ABI-related changes.
- `KERNEL_NAME` is tweaked slightly so that it can be overridden - this allows the mingw-w64 naming to be selected when compiling on Cygwin.


dra27:
Is there anything further I can do to this PR to ease its way in? 🙂

PR #304, opened 2020-10-09
Fix comment in code shown in the README.

ericonr:
I was a tad confused while looking through the readme, since the example shows setting the variable to `2` for a `1-pass computation`. I believe this makes it correct.


ericonr:
Ping

ericonr:
Ping

PR #309, opened 2020-12-03
Don't use @rpath on macOS

ryandesign:
There's no particular benefit to using `@rpath` instead of just using the absolute path like everybody else does, so just do that.

Late assignment (`=`) is used instead of immediate assignment (`:=`) because `LIBRARY_REL` isn't defined until later in the Makefile.

The user will need to specify `PREFIX` when running `make` in addition to when running `make install`. The readme is updated to reflect that.


vexy (approved):

ryandesign:
And in fact using `@rpath` can cause actual problems for users, such as https://trac.macports.org/ticket/61735. I just added argon2 support to php in MacPorts and the use of `@rpath` in the argon2 library's install name prevents the php apache module from loading, at least for that user. So this PR would fix that.

PR #321, opened 2021-06-25
fix 3 spelling mistakes

bittorf:
minor stuff, but...

compromized => compromised
securly => securely
incluse => include


veorq:
Thanks!

PR #322, opened 2021-06-28
define NATIVE_LITTLE_ENDIAN on loongarch

loongson-zn:
add the judgment of loongarch64 architecture


xen0n (changes requested):
Hi, I believe this is not needed, as `__BYTE_ORDER__` is already present on LoongArch GCC, so there's nothing to do to support LoongArch. You could simply close this.

yetist:
The following macros are built-in in the LoongArch platform:
```
$ uname -m
loongarch64
$ gcc -E -dM - < /dev/null |grep "BYTE_ORDER__\|_loongarch64"
#define __loongarch64 1
#define __BYTE_ORDER__ __ORDER_LITTLE_ENDIAN__
```

PR #332, opened 2021-12-16
Add dargon2_flutter for Flutter projects

tmthecoder:
Adds in the reference for Flutter projects, [dargon2_flutter], which differs slightly from [dargon2], which provides argon2 functionality in Dart for native devices only. 

The Flutter plugin is necessary in providing bindings for mobile & web devices which require a different method of loading the argon2 reference library.

This PR adds in the reference to the flutter plugin in the readme and clarifies the distinction between [dargon2] and [dargon2_flutter].

Let me know if I need to make any clarifications!

[dargon2]: https://github.com/tmthecoder/dargon2/blob/main/dargon2
[dargon2_flutter]: https://github.com/tmthecoder/dargon2/blob/main/dargon2_flutter

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,8 @@ Bindings are available for the following languages (make sure to read
 their documentation):
 
 * [Android (Java/Kotlin)](https://github.com/lambdapioneer/argon2kt) by [@lambdapioneer](https://github.com/lambdapioneer)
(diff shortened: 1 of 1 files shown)


tmthecoder:
@khovratovich @veorq  just a quick ping since it's been a few weeks

PR #334, opened 2021-12-23
hardened checks for numeric arguments

calestyo:
Given the security sensitive nature of the argon2 program, it should reject any
invalid numbers in its arguments.

For example, the option `-k 104u576`, which is likely just a typo and should
rather be `-k 1048576`, would be taken as 104, which is far less secure than the
probably intended 1048576.

Signed-off-by: Christoph Anton Mitterer <mail@christoph.anton.mitterer.name>


calestyo:
@khovratovich Do you know whether this is still maintained?

PR #335, opened 2021-12-23
improved the manpage

calestyo:
This includes:
- Added the note from the RFC, that the Argon2id is recommended for all
  environments.
- Added some formatting.
- The complete usage modes.
- For the 3 Argon variants, added some basic description of these, taken from
  the RFC, with the exception that Argon2i isn’t decribed as preferred.
- Mentioned which options are mutually exclusive.
- Added the missing -i and -k options.
- Some further minor textual changes.
- References to the RFC.

Signed-off-by: Christoph Anton Mitterer <mail@christoph.anton.mitterer.name>

### DIFF
--- a/man/argon2.1
+++ b/man/argon2.1
@@ -1,53 +1,87 @@
-.TH ARGON2 "1" "April 2016" "argon2 " "User Commands"
+.TH ARGON2 "1" "December 2021" "argon2 " "User Commands"
 
 .SH NAME
 argon2 \- generate argon2 hashes
 
 .SH SYNOPSIS
-.B argon2 salt
-.RB [ OPTIONS ]
+.B argon2
+\fI\,salt\/\fR [\fI\,-i\/\fR|\fI\,-d\/\fR|\fI\,-id\/\fR] [\fI\,-t iterations\/\fR] [\fI\,-m memory\/\fR | \fI\,-k memory\/\fR] [\fI\,-p parallelism\/\fR] [\fI\,-l output-length\/\fR] [\fI\,-e\/\fR|\fI\,-r\/\fR] [\fI\,-v\/\fR (\fI\,10\/\fR|\fI\,13\/\fR)]
+.br
+.B argon2
+[\fI\,-h\/\fR]
 
 .SH DESCRIPTION
 Generate Argon2 hashes from the command line.
 
-The supplied salt (the first argument to the command) must be at least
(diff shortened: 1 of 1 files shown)


calestyo:
Someone should double check especially the following:
- I took some quotes from the RFC (describing the differences of the variants). I would have said that these snippets are too small to deserve any copyright - but I'm not a lawyer.
- I also modified the portion from the RFC, that describes the i variant as preferred:
> Argon2i uses data-independent memory access, which is preferred for password hashing and password-based key derivation.

and rather added a note in the beginning that the id variant is generally preferred.

calestyo:
@dkg ... since you wrote the main part of the manpage... maybe you want to have a look at this.

dkg:
This all looks good to me.  as for the argon2i vs. argon2id choices and documentation, RFC 9106 is a little funny in that [its introduction](https://www.rfc-editor.org/rfc/rfc9106.html#name-introduction) says:

> Argon2i uses data-independent memory access, which is preferred for password hashing and password-based key derivation.

but [the recommendation section](https://www.rfc-editor.org/rfc/rfc9106.html#name-recommendations) says:

> The Argon2id variant with t=1 and 2 GiB memory is the FIRST RECOMMENDED option and is suggested as a default setting for all environments.

I think the proposed changes are fair, because arguably the sentence in the introduction is saying that 2i is "preferred…" in these contexts over 2d, due to 2d's vulnerability to side-channel attacks.

I don't have the ability to merge this change, of course, but i have no objections to it.

dkg:
This all looks good to me.  as for the argon2i vs. argon2id choices and documentation, RFC 9106 is a little funny in that [ts introduction](https://www.rfc-editor.org/rfc/rfc9106.html#name-introduction) says:

> Argon2i uses data-independent memory access, which is preferred for password hashing and password-based key derivation.

but [the recommendation section](https://www.rfc-editor.org/rfc/rfc9106.html#name-recommendations) says:

> The Argon2id variant with t=1 and 2 GiB memory is the FIRST RECOMMENDED option and is suggested as a default setting for all environments.

I think the proposed changes are fair, because arguably the sentence in the introduction is saying that 2i is "preferred…" in these contexts over 2d, due to 2d's vulnerability to side-channel attacks.

I don't have the ability to merge this change, of course, but i have no objections to it.

calestyo:
@veorq Just wondered whether this is still maintained? Last commit is already quite some time ago, and the number of MRs is rising ;-)

PR #338, opened 2021-12-30
Added explicit casts to malloc usages in order to fix compilation on C++

GlitchedPolygons:
When compiling this Argon2 lib in a C++ project, the compilation fails due to illegal conversions from `void*` (what `malloc(size_t)` returns) to `uint8_t*` - this is perfectly valid in C (and don't get me wrong, I love C just as much as everyone else) but it's a non-ignorable error in C++.

Since you put

```C
#if defined(__cplusplus)
extern "C" {
#endif
```

inside `argon2.h` I thought I'd open this pull request, because it makes no sense to hint C++ friendliness with that `extern "C"` if compilation then fails in C++ due to implicit pointer type conversions.

My idea here would be to

* Either accept this pull request and merge full C++ compatibility into Argon2 lib with minimal changes
* Or reject this pull request but then also remove the `#ifdef _cplusplus` block at the top of the header file to avoid confusion for C++ devs interested in using this library

Kindest regards, and have a happy and successful new year! :D

### DIFF
--- a/cpp.patch
+++ b/cpp.patch
@@ -1,0 +1,71 @@
+diff --git a/src/argon2.c b/src/argon2.c
+index 34da3d6..ae3abc7 100644
+--- a/src/argon2.c
++++ b/src/argon2.c
+@@ -124,7 +124,7 @@ int argon2_hash(const uint32_t t_cost, const uint32_t m_cost,
+         return ARGON2_OUTPUT_TOO_SHORT;
+     }
+ 
+-    out = malloc(hashlen);
++    out = (uint8_t *)malloc(hashlen);
+     if (!out) {
+         return ARGON2_MEMORY_ALLOCATION_ERROR;
+     }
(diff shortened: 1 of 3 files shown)


Hello71:
although I am not argon2 author, extern "C" is fully consistent and in fact mandatory. argon2 is a C library that can be *used* from C++. it is not a C++ library. almost all C++ implementations except the notoriously deficient Microsoft implementation can be configured to compile certain files in C compliant mode, and even the Microsoft implementation now has official C compliant mode. extern "C" means that the referenced functions are written in C or more precisely use the C ABI, but can be seamlessly invoked from C++.

josephlr:
@Hello71 is correct. The `#if defined(__cplusplus)` stuff means that the headers for this library are C++ compatible. The fact that the implementation files for this library end in `.c` indicate that _the library itself_ should be built with a C compiler not a C++ compiler.

I think this PR can be closed unless we think that adding the explicit casts is good **C** style. C++ compatibility for the `.c` files should be a non-goal in my opinion.

PR #377, opened 2024-07-19
Make fallthroughs explicit encoding.c

r-barnes:
This allows `-Wimplicit-fallthrough` to pass on LLVM.


Leont:
I don't think that's portable, so you need to wrap it in some defines.

r-barnes:
@Leont - I've added a macro for portability.

